PR #22, opened 2015-09-06
Fix #21 by upgrading gulp-typescript

ibrahima:
See https://github.com/Microsoft/TypeScript/issues/4565

PR #25, opened 2015-09-06
Removed aboslute path import and unnecesarry circular import

jimmyhmiller:
Once I upgrade gulp as per #22, I noticed an import failure. This fixes that problem and makes the app actually work on my machine.

Thanks for the awesome project.


Fzzr:
This addressed #26 for me.

shockone:
Thank you for the contribution, @jimmyhmiller. I would appreciated if you or anybody else helped me to develop this project.

PR #37, opened 2015-09-06
Add license file

jacwah:
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

Just writing "License: The MIT License" is a) not permitted by said license and b) not legally sound. See [this StackEchange question](http://opensource.stackexchange.com/q/176/478) for more info.


shockone:
@jacwah, thank you.

PR #53, opened 2015-09-07
Update README.md

zedtux:
Add installation of Gulp.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,14 @@ capabilities of contemporary interfaces in contrast to most of the terminal emul
 Right now only OS X is supported, but we plan to make it work on all the major platforms.
 
 ```bash
-npm run install-all
+$ npm run install-all
+$ npm install --global gulp
 ```
 
 ### Run
 
 ```bash
-gulp
+$ gulp
 ```
 
 ### Package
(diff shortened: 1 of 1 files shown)


G07cha:
Gulp exists in devDependencies, so better just move gulp to 'dependencies' object

zedtux:
Thank you @G07cha. Should I move all the `gulp-*` thing or just `"gulp": "3.9.0",` ?

shockone:
@G07cha I'm not sure what's the point of moving it into dependencies. Could you elaborate, please?

G07cha:
Yep, I've just noticed that for production build we don't need gulp at all, so this pull request useless for regular users, but anyway, need to make sure that gulp tasks don't creates files nessesary for electron-package.

shockone:
That's true, but I think the initial issue was that building the terminal is invoked by the `gulp` command, which is not necessarily in the path. As an idea, we can create an `npm` script to start `gulp`. This way it will surely appear in the path.

G07cha:
Hence I think we can close this request and for feature need to create .app build for production and guide in README mark as build for development

shockone:
@zedtux, now you can start the build with `npm start` instead of `gulp`. 

@G07cha, I'll create an issue about the build.

zedtux:
Thank you @shockone for this. I'll give it a try.

Issue #51, opened 2015-09-07
No need to add node_modules to path


markogresak:
In [`package.json`](https://github.com/shockone/black-screen/blob/master/package.json#L60-L63), you're adding `node_modules/.bin` to path. This is not required, since `npm run [script]` will automatically try to find the binaries in `.bin` folder.

iiegor:
+1 

I think the entire ``PATH=node_modules/.bin:$PATH rm -rf node_modules`` is unnecesary and also makes windows users can't be able to run the script.

markogresak:
Using `rm -rf` is the least of problems that make it unusable on Windows. If you look at the source, it's assuming *nix architecture so making it work on windows would require major refactoring.

PR #54, opened 2015-09-07
Removed node_modules/.bin path

G07cha:
Based on #51 conversation I've remove `node_modules/.bin` from package.json, have run affected scripts and seem like everything work fine


shockone:
Wow, I didn't know that. Merging.

PR #75, opened 2015-09-08
Add overlay-scrollbars

G07cha:
Now scrollbars are hidden by default and appear only on scrolling also they have dark color what prettier that default OS X

**Before:**

screen shot 2015-09-08 at 2 13 32 pm

**After:**
screen shot 2015-09-08 at 3 09 41 pm

**On scrolling:**

screen shot 2015-09-08 at 2 11 25 pm
Here drop-down list hidden behind the line becouse of strange screenshot system

### DIFF
--- a/src/main/main.js
+++ b/src/main/main.js
@@ -13,6 +13,9 @@ app.on('ready', function () {
     mainWindow = new BrowserWindow({
         width: size.width,
         height: size.height,
+		'web-preferences': {
+			'overlay-scrollbars': true
+		},
         resizable: true
     });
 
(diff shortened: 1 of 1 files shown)


wngravette:
looking good! :+1:

iiegor:
+1

src/main/main.js
```diff
@@ -13,6 +13,9 @@ app.on('ready', function () {
     mainWindow = new BrowserWindow({
         width: size.width,
         height: size.height,
+		'web-preferences': {
+			'overlay-scrollbars': true
+		},
```
shockone: Could you format it to have the same indentation?

PR #88, opened 2015-09-08
Use pty.js instead of child_pty.

shockone:
* Use pty.js instead of child_pty.
* Simplify the build command.

### DIFF
--- a/src/Invocation.ts
+++ b/src/Invocation.ts
@@ -1,6 +1,6 @@
 /// <reference path="references.ts" />
 
-var child_pty = require('child_pty');
+var pty = require('pty.js');
 import child_process = require('child_process');
 import _ = require('lodash');
 import React = require('react');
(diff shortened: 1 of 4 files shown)


shockone:
@black-screen/team please review.

G07cha:
I planned to test tomorrow on the OS X

shockone:
I imagined a code review to be a review of only the code quality, irrelatively to whether it works. :)

G07cha:
Sure I've looked on a code and it's seems fine so I'll do my best, test if it's working.

src/Invocation.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference path="references.ts" />
 
-var child_pty = require('child_pty');
+var pty = require('pty.js');
```
RobertoUa: why not import?

shockone: There is no type definition for this module. It won't compile with `import`.

RobertoUa: is it hard to add the type definition with one 'spawn' method?

shockone: Not really, similar to this https://github.com/black-screen/black-screen/blob/master/dts/AnsiParser.d.ts. Will you? :)

RobertoUa:
awesome!

shockone:
@RobertoUa, I know, right?

PR #89, opened 2015-09-08
Move bower to postinstall.

shockone:
* Don't use `rm` because it doesn't work on Windows.
* Move the `bower` commands to the `postinstall` hook since it might not be installed yet.

PR #91, opened 2015-09-08
Include prerequisites in README install instructions.

xtoddx:
This is a small change to README.md that makes mention of how to install node and use npm to install bower globally. These are required to complete the first `npm install` step of the installation instructions.

Since on most machines this will already be done, and is not specific to black-screen, I created a `prerequistes` section in the README so most people can just skip over it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,19 @@ As opposed to most of the terminal emulators, it exploits the capabilities of co
 
 Note: Only OS X is currently supported, but we plan to make Black Screen work on all the major platforms.
 
+###### Prerequisites
+You'll need a working node installation, and bower installed globally.
+Use [homebrew](http://brew.sh/) to install node.
+```bash
+brew install node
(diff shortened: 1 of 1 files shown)


G07cha:
Bower is in dependencies already so it must be installed with npm install command.

G07cha:
Also I like compact model, like [this one](https://github.com/ekonstantinidis/gitify#prerequisites).

shockone:
@xtoddx, I agree with @G07cha. We do expect some level of knowledge from our users (prerequisites, huh?). I think it's hard to find a developer who doesn't know how to install node.js.

xtoddx:
Maybe we were racing each other :racehorse:  #89 landed after my fork, it fixes the use of bower in a `preinstall` script.

shockone:
@xtoddx that's true. Thank you for help in any case.

Issue #105, opened 2015-09-09
CodeQuality: tslint


joshtoo:
Enable tslint rules and gulp automation

PR #106, opened 2015-09-09
Issue #101 JSHint Implementation

### LINKED ISSUE #101: CodeQuality: JSHint
Enable JSHint rules and gulp automation

joshtoo:
Fixes #101 

* Most enforcing options enabled
* No relaxing options enabled
* No deprecated options listed


shockone:
@joshtoo, I'm not sure you noticed, but we write TypeScript. There are a couple of views in JSX, but they'll be ported to TS as well.

joshtoo:
@shockone TypeScript is covered by #105 

JS Code quality stuff if pretty boiler plate at this point, not a lot of time spent here to cover the JSX views.

RobertoUa:
If choose between jshint and eslint I would pick eslint.

shockone:
@RobertoUa, could you explain why?

joshtoo:
@RobertoUa - I went with JSHint in this case because of the very limited number of .js files and the quick implementation. If we were adding a bunch of .js I would have probably gone with eslint. In the end it really doesn't matter to me, i'm more than happy to go with eslint. It certainly wouldn't take long to swap them.
@shockone - To answer your question, eslint is a superior tool. Extensible / Plugin support more options and better ES6 support. The trade off is configuration time and it tends to run a bit slower.

RobertoUa:
> configuration time 

It's usually just downloading some popular configuration from the internet. Because usually developers follow code style from another developers.

>JSHint won't work with babel.

PR #109, opened 2015-09-09
#108 Open app from dock

### LINKED ISSUE #108: Reopen app from dock
On OS X, when closing app by tapping cross button app moving to a background but can't started again from dock.

G07cha:
Fixes #108 

Current implementation:
- On opening new window app will be showed only after web content will be loaded(prevent appearing white screen)
- On reopening app from dock(OS X) will be started new app
- Window size set to 700x450 as suggested @iiegor

[Screencast](https://vimeo.com/138723730)

### DIFF
--- a/src/main/main.js
+++ b/src/main/main.js
@@ -26,6 +26,7 @@ function createWindow() {
 	var userPrefsPath = path.join(app.getDataPath(), "preferences.json");
 	var data;
 	
+	//Reading user's preferences file
 	try {
 		data = JSON.parse(fs.readFileSync(userPrefsPath, 'utf8'));
 	} catch(e) {
@@ -43,6 +44,7 @@ function createWindow() {
 	});
 	
 	if(data) {
+		//Applying user's preferences
 		if(data.maximized) {
 			window.maximize();
 		} else {
@@ -54,6 +56,7 @@ function createWindow() {
 	menu.setMenu(app, mainWindow);
 	
 	window.on('close', function () {
+		//Remembering window size and position before exit
 		var data = {
 			bounds: mainWindow.getBounds(),
 			maximized: window.isMaximized()
--- a/src/main/main.js
+++ b/src/main/main.js
@@ -7,19 +7,41 @@ process.env.PATH += ':/usr/local/bin';
 var mainWindow;
 
 app.on('ready', function () {
-
-    var atomScreen = require('screen');
-    var size = atomScreen.getPrimaryDisplay().workAreaSize;
-    mainWindow = new BrowserWindow({
-        width: size.width,
-        height: size.height,
-	'web-preferences': {
-		'overlay-scrollbars': true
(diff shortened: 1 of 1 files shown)


shockone:
I think we should remember the window size.

RobertoUa:
@shockone yes remember the size by default. But also add a settings window where user can enable/disable it. And also add an option to "aways start with maximized window"

shockone:
I guess settings should be a separate task.

RobertoUa:
@shockone yes, of course.

G07cha:
Will do.

G07cha:
Done, also partial-implemented 'Start maximized' feature(OS X don't allow start applications in full screen by default as i know)

src/main/main.js
```diff
@@ -7,19 +7,68 @@ process.env.PATH += ':/usr/local/bin';
...
-    mainWindow.focus();
-    menu.setMenu(app, mainWindow);
+app.on('activate-with-no-open-windows', function () {
+	mainWindow = createWindow();
 });
+
+function createWindow() {
+	var path = require("path");
+	var fs = require("fs");
+	var userPrefsPath = path.join(app.getDataPath(), "preferences.json");
```
shockone: We already have serialization setup. Could you try to use it? https://github.com/black-screen/black-screen/blob/master/src/Terminal.ts#L98

src/main/main.js
```diff
@@ -7,19 +7,68 @@ process.env.PATH += ':/usr/local/bin';
...
+	menu.setMenu(app, mainWindow);
+	
+	window.on('close', function () {
+		//Remembering window size and position before exit
+		var data = {
+			bounds: mainWindow.getBounds(),
+			maximized: window.isMaximized()
+		};
+		
+		fs.writeFileSync(userPrefsPath, JSON.stringify(data));
```
shockone: I think we can do that on resize.

G07cha: Not necessary, if user closing window he expects that on opening next time it will have same size.

shockone: Okay.

src/main/main.js
G07cha: I had no experience with typescript but I'll try.

joshtoo:
@RobertoUa - All of the linting support for .js & .ts will be in shortly.

shockone:
@G07cha, how about leaving the window dimensions untouched in this PR and dealing with them separately? After all, the title says "Open app from dock".

G07cha:
> shockone
> I think we should remember the window size.

I just did what you said.

shockone:
I still believe it's the right thing to do, but maybe we should separate these issues. Although, you decide.

G07cha:
Merged commit with implemented reopening from dock functionality, 'Save window bounds' feature will be in separate PR, closing this one.

Issue #15, opened 2015-07-11
Set up Docker


shockone:
* Create a container which would have everything required for running specs and create an npm script `npm test` (see `package.json` for other scripts) to run those specs.

* Create a container which would have everything set up for new contributors to start working on the project. The goal is to make setup as easy as possible.

jonadev95:
- So, I got my local copy of the repo, I mount it into the container and let the container do the test?
- That means all required tools and the up-to-date repo?

shockone:
@jonadev95, that's the plan.

jonadev95:
A little addition to the second point, so they should work inside the container?
I'll do it if you want :+1:

shockone:
I don't have enough knowledge to answer that, but the goal is to make the development easy. I guess, it would be harder to work inside the container, right?

Please collaborate with @kblanks, he's also working on that.

kblanks:
@jonadev95 I'm cool if you want to do this yourself, I'll find another issue to work on :+1:

jonadev95:
@shockone depends, but a problem would be that it's possible to loose all of your work when accidentilly deleting the container. So better keep the code outside of it.

@kblanks thanks man :+1:   

I will start working on it.

joshtoo:
@jonadev95 - Once we have an approved container, I will add the docker build / publish to Travis ( See #63 )

jonadev95:
@joshtoo great! 
So someone will set up an account on hub.docker.com?

G07cha:
@joshtoo, as I know Travis already run builds in dockers.

jonadev95:
@G07cha If I understand it the right way, he means that he will add a job to build the dockerfiles and push them automatically to the docker hub repo

G07cha:
Okay, I get it, but first of all we need to make black-screen stable build for Linux becouse all docker images are linux-based.

jonadev95:
@G07cha I've tried it on my linux pc (Because I didn't see note that it's only for OS X ), and with gulp it works very well.

But if you think it's better to wait until it's stable I will write it based on the current state, but will wait for the stable release before opening a pull request.

ruffsl:
This looks awesome. I've been playing with this project on Ubuntu, and aside from some "JavaScript error occurred in the main process" error popups, this looks promising. Have you thought of running the GUI interface within the container, using the host's Xserver with Unix sockets?

jonadev95:
[...]

PR #117, opened 2015-09-09
Docker Image for testing

sauercrowd:
#15 It would be great if I could get any feedback for the Image before closing the PR.
I've made two scripts, build.sh and run.sh, to make it easy for people who are not familiar with docker yet. (description in docker/README.md)

The image is quite big, ~1GB on my HD. So when executing build.sh keep in mind that it will take some time.


shockone:
Hey, @jonadev95. Thank you for your efforts. I'll look into it as soon as possible.

docker/testrunner/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+cd /black-screen
+
+if [[ ! -e /.firstrun ]] && ( [[ ! -d node_modules ]] || [[ -n $FORCE ]] )
```
shockone: `firstrun` sounds like it should be present only on the first run, but it's the opposite.

docker/testrunner/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+	npm install -g selenium-standalone
+	selenium-standalone install
+	chown -R testrunner:users .
+	sudo -u testrunner echo '{ "interactive": false }' > /home/testrunner/.bowerrc
+	sudo -u testrunner npm install
+	touch /.firstrun
+fi
+
+selenium-standalone start &
+sleep 5
```
shockone: I guess there is no better way to wait for it, right?

docker/testrunner/run.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
```
shockone: Can we make it dependent on `build.sh`? Meaning that if I execute `run.sh` and there is no container yet, it would build it.

docker/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+It doesn't contain the source code, so you need to mount the git repo at **/black-screen**
+
+##### Usage
+
+**Change into the testrunner directory**
+
+- build the container with ```docker build -t black-screen_testrunner . ```
+- first startup: ```docker run --name bs-testrunner -v `pwd`/black-screen:/black-screen black-screen_testrunner ``` (Replace \`pwd\`/black-screen with your absolute path to the repo)
+- restart of tests: ```docker restart bs-testrunner; docker attach bs-testrunner```
```
shockone: I think this section is unnecessary. As long as people can run the shell scripts, they shouldn't care about the implementation.

docker/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+For all tasks are scripts available:
+ - build: `./build.sh`
+ - run/re-run `./run.sh`
+
+Optional parameters for `./run.sh`:
+- [-f]  Force to rebuild the container, for e.g. updating of node dependencies (normally restart the container if there's one available)
```
shockone: In the future we'd like to automatically understand when to rebuild.

docker/testrunner/entrypoint.sh
jonadev95: So `  ! -e /.firstun  ` is true when firstrun file is present?

docker/testrunner/entrypoint.sh
jonadev95: I sadly didn't find a better way, but if anyone has an idea let me know!

docker/testrunner/entrypoint.sh
shockone: It works correctly, I'm talking about the name. It reads like "is it the first run?" for me. Although, who cares?

shockone:
Thank you for the contribution, @jonadev95. Since You saved me a lot of time!

docker/testrunner/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+cd /black-screen
+
+if [[ ! -e /.firstrun ]] && ( [[ ! -d node_modules ]] || [[ -n $FORCE ]] )
+	then
+	npm install -g selenium-standalone
+	selenium-standalone install
+	chown -R testrunner:users .
+	sudo -u testrunner echo '{ "interactive": false }' > /home/testrunner/.bowerrc
+	sudo -u testrunner npm install
```
shockone: I'm wondering if this command changes the modules on my local machine as well?

shockone:
Do the modules build for you?

docker/testrunner/entrypoint.sh
jonadev95: Currently it does. If that isn't the way it should be let me know.

jonadev95:
@shockone The day before yesterday it does... I haven't tested it today, but on my local linux machine I'm getting errors when npm tries to install typescript, is that what you're experiencing?

docker/testrunner/entrypoint.sh
shockone: Is it possible to install the modules somewhere else? I use OS X, and it would make me recompile the local modules every time I use the container.

shockone:
@jonadev95, no, for me pty.js doesn't compile.

jonadev95:
@shockone branches... you're completely right, when trying to build it on master pty.js is not building. Then I switched to develop branch (Just to try it out) and then I get this typescript error.

jonadev95:
@shockone tests are passing again

docker/testrunner/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+	touch /.firstrun
+fi
+
+selenium-standalone start &
+sleep 5
+xvfb-run npm run test
```
shockone: Can be just `+xvfb-run npm test`

shockone:
@jonadev95, just remembered: one of the points of a container for testing was that if I accidentally execute `rm -rf *` in them (because it's a terminal, you know, it executes commands), I wouldn't have to worry about my files.

jonadev95:
@shockone Little Summary:

There's a new option for run.sh, -b, to force a rebuild of the whole image.
If the user doesn't pass a path to run.sh, the container will clone the latest state from github.
Before building it, the container copies the whole directory which is mounted and execute npm install and npm test there, so nothing on the host is affected

shockone:
@jonadev95, awesome. Please let me know whether it's ready for testing.

jonadev95:
@shockone I will run a few tests to make sure all possible options are working as expected, but after that it's ready :+1: I will write a comment here when I'm finished.

jonadev95:
@shockone I'm done. I've tried to remove files after before copying the sources after the first run, but that brings the problem that node_modules, bower_modules,... are removed as well. As a workaround I think it's the best to not remove old files, instead if the build fails because of old files, the user should rebuild the container with -f (I hope the problem is clear). What do you think about that?

shockone:
How about creating the `node_modules` folder in a different place and making a symlink in the project directory?

jonadev95:
@shockone tried different options of symlinking, but npm doesn't seem to like that...
the solution I've used is to copy the node and bower modules to a third location and copy it back again after beginning from the second run.
It's not brilliantly effective, but at least it works...

shockone:
@jonadev95, hey, sorry I haven't been answering, I'm really busy these days. I guess it's ready to merge, right?

jonadev95:
@shockone No Problem, great you're back :-)
Right :+1:

docker/testrunner/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update
```
bpinto: It is recommended to run both `apt-get` commands in a single `RUN` statement:

```
RUN apt-get update && apt-get install -y \
        openjdk-7-jdk \
        sudo \
        xorg \
        xserver-xorg-video-dummy \
        xvfb
```

Reference: https://docs.docker.com/engine/articles/dockerfile_best-practices/#apt-get

docker/testrunner/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+RUN apt-get install -y sudo openjdk-7-jdk xorg xvfb xserver-xorg-video-dummy
+RUN useradd -m -G users testrunner
+ADD entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
```
bpinto: Is this necessary? I thought docker would keep the permissions of a file when using the `ADD` instruction.

hickscorp: Maybe that user is running Windows, and mounting the source in his container?

bpinto: Still this `chmod` would be executed when building the image and not inside the container with the mounted folder.

docker/testrunner/Dockerfile
sauercrowd: I'm sorry for such an late answer, but you're completely right, it should be a one liner.

docker/testrunner/Dockerfile
sauercrowd: Again, sorry for such a late answer.
Regarding the permissions, I wanted to make sure the file got the right permissions.
I wasn't aware it would be useless since it is executed during the building process.

Issue #112, opened 2015-09-09
Extend Background on Multi-line Input 


draughoncollin:
The lighter grey that serves as the background for your command-line input should have multi-line functionality:
![screen shot 2015-09-09 at 6 40 55 am](https://cloud.githubusercontent.com/assets/6424040/9760783/c7b7d17c-56bd-11e5-846a-40b9c1497d16.png)

draughoncollin:
Probably a related issue but putting in multiple multi-line inputs causes an overwrite of the next line:
![screen shot 2015-09-09 at 7 13 58 am](https://cloud.githubusercontent.com/assets/6424040/9761457/7743ac20-56c2-11e5-9112-8c552d9a3a8b.png)

joshtoo:
True Story - OSX 10.11 Beta ( Not that this is going to be OS specific )
<img width="953" alt="screen shot 2015-09-09 at 7 17 23 am" src="https://cloud.githubusercontent.com/assets/8420490/9761522/df018a26-56c2-11e5-8166-1477af777e19.png">

shockone:
Fixed. ![](https://dl.dropboxusercontent.com/spa/dlqheu39w0arg9q/a771n893.png)

PR #118, opened 2015-09-09
Fixes Multi-line Input | issue #112

egladman:
##### this addresses issue #112 HOWEVER this also introduces two minor issues
- the cursor is hidden behind the class: prompt-decoration. Once text is inputted  the cursor becomes visible. 
- the class: autocomplete becomes hidden once it exceeds the viewport

### DIFF
--- a/stylesheets/main.scss
+++ b/stylesheets/main.scss
@@ -2,7 +2,7 @@
 
 $fa-font-path: "../node_modules/font-awesome/fonts";
 $font-size: 14px;
-$prompt-padding: 5px;
+$prompt-padding: 0px;
 $default-shadow: 0 2px 8px 1px rgba(0, 0, 0, 0.3);
 
 @import '../node_modules/font-awesome/scss/font-awesome';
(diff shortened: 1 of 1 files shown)


shockone:
Hey, @egladman. Sorry for a long delay. I'm afraid I have little time now, but I'll test it as soon as possible.

egladman:
@shockone thanks for the update. I will try to address the two minor issues that i unintenially introduced.

shockone:
@egladman, how do you even make a multi line prompt?

egladman:
@shockone short answer: type gibberish. Currently if an excessively long string(without spaces) is entered the line is not wrapped and the overflow is hidden.  I think the overflow should be horizontally scrollable. If you share your thoughts on the matter I can revise my commit.

shockone:
@egladman, I did that in your branch, and it is still not wrapped.

pixeldesu:
If you want to wrap long lines with CSS you should probably attempt using `word-break: break-all;` for the  input.

shockone:
Fixed with CSS Grid Layout.

PR #129, opened 2015-09-12
Standalone windows patch

iiegor:


src/CommandExecutor.ts
```diff
@@ -51,6 +51,12 @@ class BuiltInCommandExecutionStrategy extends CommandExecutionStrategy {
 class SystemFileExecutionStrategy extends CommandExecutionStrategy {
     startExecution() {
         return new Promise((resolve, reject) => {
```
shockone: The purpose of the strategy classes is to [replace conditions with polymorphism](http://www.refactoring.com/catalog/replaceConditionalWithPolymorphism.html). Please create a separate strategy for Windows.

src/Utils.ts
```diff
@@ -123,6 +123,20 @@ export default class Utils {
...
 
+    static getCmdPath(): string {
+        if (process.env.comspec) {
+            return process.env.comspec;
+        }
+        else if (process.env.SystemRoot) {
+            return Path.join(process.env.SystemRoot, 'System32', 'cmd.exe');
+        }
+        else return 'cmd.exe';
+    }
```
shockone: Since it's Windows-only, please move it to the strategy class mentioned above.

src/Aliases.ts
```diff
@@ -14,6 +14,8 @@ export default class Aliases {
     }
 
     static importAliasesFrom(shellName: string): void {
+        if (process.platform === 'win32') return;
```
shockone: How about creating an enum with platforms in `Enums.ts` and a helper function `Utils.isWindows`?

src/Buffer.ts
```diff
@@ -34,6 +34,9 @@ export default class Buffer extends events.EventEmitter {
         var current = {attributes: <i.Attributes>null, text: ''};
         var cursorPosition = cursor.getPosition();
 
+        /* Quick fix for the strange character at the end (https://github.com/black-screen/black-screen/pull/56#issuecomment-139653723) */
+        if (process.platform === 'win32' && row.length < 1) return;
```
shockone: Triple equal is useless in TypeScript. In JS it compares with types, but, you know, **Type**Script compares types out of the box. :)

src/CommandExecutor.ts
```diff
@@ -51,6 +51,12 @@ class BuiltInCommandExecutionStrategy extends CommandExecutionStrategy {
...
+            if (process.platform === 'win32') {
+                this.args.unshift(this.command);
+                this.args = ['/s', '/c', this.args.join(' ')];
```
shockone: Have you tried executing a command with multiple parameters? I'm not sure it works as expected.

src/CommandExecutor.ts
```diff
@@ -51,6 +51,12 @@ class BuiltInCommandExecutionStrategy extends CommandExecutionStrategy {
...
+                this.command = Utils.getCmdPath();
```
shockone: Please do this in the constructor.

src/CommandExecutor.ts
```diff
@@ -61,7 +67,8 @@ class SystemFileExecutionStrategy extends CommandExecutionStrategy {
 
             this.invocation.command.stdout.on('data', (data: string) => this.invocation.parser.parse(data.toString()));
             this.invocation.command.on('exit', (code: number) => {
-                if (code === 0) {
+                /* In windows there is no code returned (null) so instead of comparing to 0 we check if its 0 or null with ! */
+                if (!code) {
```
shockone: How about converting to an integer first. The reason is only readability: `!code` doesn't look like a success case for me.

src/Terminal.ts
```diff
@@ -37,14 +37,17 @@ export default class Terminal extends events.EventEmitter {
...
-            if (app.dock) {
-                app.dock.bounce('informational');
-            }
-            this.createInvocation();
-        }).once('working-directory-changed', (newWorkingDirectory: string) =>
-            this.setCurrentDirectory(newWorkingDirectory)
-        );
+
+        invocation
+            .once('clear', _ => this.clearInvocations())
```
shockone: We'll remove events in the future, so try to avoid them. Although, I'll accept this PR even with the new event, because it might be hard to implement it in a different way.

src/Utils.ts
```diff
@@ -123,6 +123,20 @@ export default class Utils {
...
+
+    static getHomeDirectory(): string {
+        return process.env[(process.platform === 'win32') ? 'USERPROFILE' : 'HOME'];
```
shockone: Remember about `===` here and everywhere else.

src/Buffer.ts
etiktin: It's not a mistake to use strict equality (triple equals) in TypeScript. When you read a piece of code that contains strict equality it's obvious to the reader that no type coercion will happen there. TypeScript transpiles equality operators as is to the JavaScript code, so the meaning is still clear in js.
Also if you take a look at the official [TypeScript samples](http://www.typescriptlang.org/Samples), you will see they almost always use strict equality in their ts files (e.g. [raytracer.ts](https://github.com/Microsoft/TypeScriptSamples/blob/master/raytracer/raytracer.ts)).

shockone: You're right, my bad.

src/CommandExecutor.ts
shockone: I've refactored the way we choose a strategy to make it easier to extend. Sorry if it created conflicts.

iiegor:
Done, check now @shockone

src/CommandExecutor.ts
```diff
@@ -63,6 +64,47 @@ class SystemFileExecutionStrategy extends CommandExecutionStrategy {
     }
 }
 
+class WindowsSystemFileExecutionStrategy extends CommandExecutionStrategy {
+    static canExecute(command: string): Promise<boolean> {
+        return new Promise(resolve => Utils.getExecutablesInPaths().then(executables => resolve(_.include(executables, command))));
```
shockone: I guess this should be 
```typescript
return new Promise(resolve => resolve(Utils.isWindows()));
```

src/CommandExecutor.ts
```diff
@@ -63,6 +64,47 @@ class SystemFileExecutionStrategy extends CommandExecutionStrategy {
...
+    }
+
+    startExecution() {
+        return new Promise((resolve, reject) => {
+            this.args.unshift(this.command);
```
shockone: You can call `Prompt.getWholeCommand()` instead of building it from parts.

src/CommandExecutor.ts
```diff
@@ -76,14 +118,17 @@ class NullExecutionStrategy extends CommandExecutionStrategy {
 export default class CommandExecutor {
     private static executors = [
         BuiltInCommandExecutionStrategy,
-        SystemFileExecutionStrategy,
+        process.platform === 'win32' ? WindowsSystemFileExecutionStrategy : SystemFileExecutionStrategy
```
shockone: Just add two of them unconditionally, maybe we need to rename the old one to `UnixSystem...`. They should figure out if they're applicable.

src/Terminal.ts
shockone: I haven't noticed the previous time: what is `'clear'`? You must have accidentally added it to this PR.

src/Utils.ts
```diff
@@ -123,6 +123,10 @@ export default class Utils {
         });
     }
 
+    static getHomeDirectory(): string {
```
shockone: I can see you use the `get` syntax in the other place. Maybe we should do it here as well?

src/CommandExecutor.ts
```diff
@@ -63,6 +64,47 @@ class SystemFileExecutionStrategy extends CommandExecutionStrategy {
     }
 }
 
+class WindowsSystemFileExecutionStrategy extends CommandExecutionStrategy {
```
shockone: Please create the hierarchy 
```
CommandExecutionStrategy
    SystemFileExecutionStrategy
        WindowsSystemFileExecutionStrategy
        UnixSystemFileExecutionStrategy
```

and extract the common logic to `SystemFileExecutionStrategy`.

src/Aliases.ts
```diff
@@ -14,6 +14,8 @@ export default class Aliases {
     }
 
     private static importAliases(shellName: string = process.env.SHELL): void {
+        if (process.platform === 'win32') return;
```
shockone: Have you noticed my comment about an enum? https://github.com/black-screen/black-screen/pull/129#discussion_r39341819

package.json
```diff
@@ -29,7 +29,7 @@
     "lodash": "3.10.1",
     "node-ansiparser": "2.0.2",
     "octicons": "2.4.1",
-    "pty.js": "0.3.0",
+    "ptyw.js": "^0.3.2",
```
shockone: Could you please just change the version to point to your master on Github instead of changing the name?

iiegor: @shockone If I point to http://github.com/iiegor/pty.js, the git submodule of the repo doesn't initialize, so node-gyp rebuild fails :confused:

src/Terminal.ts
iiegor: It clears the invocations, like pressing CTRL + L but typing the command clear.

iiegor:
Can you fix https://github.com/black-screen/black-screen/pull/129#discussion-diff-39346906 for me? I can't find a good way to do that and also I find this like there are a lot of classes that are not specially necessary.

Regarding the enum suggestion (https://github.com/black-screen/black-screen/pull/129#discussion-diff-39346910), why would you like to do something like that? I mean, I see this also like unecessary code for a thing that can be done with a single function.

src/Terminal.ts
shockone: I understand, but there  is no command typing part of the story in this PR, right? And shouldn't be for that matter, because it's not about Windows support.

package.json
shockone: Understood, let's leave it as is.

shockone:
@iiegor, I'll do that in the evening.

Regarding enums, maybe it's just me, but I thought it's a common practice to avoid string literals comparison. With enums you always know the list of the values that exist instead of guessing whether it's 'win', 'Win', 'windows', or something else. And the program doesn't compile if you misspell the word. Anyway, it's not important, if you don't agree, I don't insist.

shockone:
@iiegor I figured I can merge and then change the hierarchy on my own.

Also, I want you to know that this pull request is the most valuable contribution so far. Thank you very much. May I ask if you plan to continue helping the project?

iiegor:
@shockone Yes I will, love this project :smile:

shockone:
@iiegor we need to figure out how to make that happen for them automatically.

iiegor:
@shockone hmm, I'm thinking on something like atom uses, [see here](https://github.com/atom/atom/tree/master/script) and move all the scripts there, so then we can check for the platform easily and apply the right commands for each platform.

onbjerg:
@iiegor Does not work on Windows 10. 

```
redacted@redacted /d/projects/bls/node_modules/ptyw.js (master)
$ HOME=~/.electron-gyp node-gyp rebuild --msvs_version="2013" --target=0.30.0 --arch=ia32 --python=/d/Python/2.7/python
.exe --dist-url=https://atom.io/download/atom-shell
gyp info it worked if it ends with ok
gyp info using node-gyp@3.0.3
gyp info using node@0.12.7 | win32 | ia32
gyp info spawn d:/Python/2.7/python.exe
gyp info spawn args [ 'c:\\Users\\nordbjerg\\AppData\\Roaming\\npm\\node_modules\\node-gyp\\gyp\\gyp_main.py',
gyp info spawn args   'binding.gyp',
gyp info spawn args   '-f',
gyp info spawn args   'msvs',
gyp info spawn args   '-G',
gyp info spawn args   'msvs_version=2013',
gyp info spawn args   '-I',
gyp info spawn args   'd:\\projects\\bls\\node_modules\\ptyw.js\\build\\config.gypi',
gyp info spawn args   '-I',
gyp info spawn args   'c:\\Users\\nordbjerg\\AppData\\Roaming\\npm\\node_modules\\node-gyp\\addon.gypi',
gyp info spawn args   '-I',
gyp info spawn args   'c:\\Users\\nordbjerg\\.electron-gyp\\.node-gyp\\0.30.0\\common.gypi',
gyp info spawn args   '-Dlibrary=shared_library',
gyp info spawn args   '-Dvisibility=default',
gyp info spawn args   '-Dnode_root_dir=c:\\Users\\nordbjerg\\.electron-gyp\\.node-gyp\\0.30.0',
gyp info spawn args   '-Dnode_gyp_dir=c:\\Users\\nordbjerg\\AppData\\Roaming\\npm\\node_modules\\node-gyp',
gyp info spawn args   '-Dnode_lib_file=node.lib',
gyp info spawn args   '-Dmodule_root_dir=d:\\projects\\bls\\node_modules\\ptyw.js',
gyp info spawn args   '--depth=.',
gyp info spawn args   '--no-parallel',
gyp info spawn args   '--generator-output',
gyp info spawn args   'd:\\projects\\bls\\node_modules\\ptyw.js\\build',
gyp info spawn args   '-Goutput_dir=.' ]
gyp info spawn C:\Windows\Microsoft.NET\Framework\v4.0.30319\msbuild.exe
gyp info spawn args [ 'build/binding.sln',
gyp info spawn args   '/clp:Verbosity=minimal',
gyp info spawn args   '/nologo',
gyp info spawn args   '/p:Configuration=Release;Platform=Win32' ]
Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
d:\projects\bls\node_modules\ptyw.js\build\deps\winpty\winpty-agent.vcxproj(20,3): error MSB4019: The imported project
"d:\Microsoft.Cpp.Default.props" was not found. Confirm that the path in the <Import> declaration is correct, and that
the file exists on disk.
d:\projects\bls\node_modules\ptyw.js\build\deps\winpty\winpty.vcxproj(20,3): error MSB4019: The imported project "d:\Mi
crosoft.Cpp.Default.props" was not found. Confirm that the path in the <Import> declaration is correct, and that the fi
le exists on disk.
gyp ERR! build error
gyp ERR! stack Error: `C:\Windows\Microsoft.NET\Framework\v4.0.30319\msbuild.exe` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onExit (c:\Users\nordbjerg\AppData\Roaming\npm\node_modules\node-gyp\lib\build.js:270:23)
gyp ERR! stack     at ChildProcess.emit (events.js:110:17)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:1074:12)
gyp ERR! System Windows_NT 6.3.9600
gyp ERR! command "d:\\Program Files (x86)\\nodejs\\node.exe" "c:\\Users\\nordbjerg\\AppData\\Roaming\\npm\\node_modules\\node-gyp\\bin\\node-gyp.js" "rebuild" "--msvs_version=2013" "--target=0.30.0" "--arch=ia32" "--python=d:/Python/2.7/python.exe" "--dist-url=https://atom.io/download/atom-shell"
gyp ERR! cwd d:\projects\bls\node_modules\ptyw.js
gyp ERR! node -v v0.12.7
gyp ERR! node-gyp -v v3.0.3
gyp ERR! not ok
```

MarkTiedemann:
@iiegor @onbjerg Same [error](https://dl.dropboxusercontent.com/u/115291845/public/black-screen-error-m2b4019.png) here. 

Note: To get it working until that point, I also had to install node-gyp and downgrade my python version to 2.7.10 since node-gyp only supports python versions between 2.5.0 and 3.0.0.

shockone:
@iiegor, well, if we have to. But before that, could you check if you can make `electron-rebuild` work with that parameter? It's much simpler that way.

onbjerg:
@MarkTiedemann Same thing here.

iiegor:
@onbjerg *d:\Microsoft.Cpp.Default.props* is missing, download Visual Studio 2013 and try again, please.

@shockone Running: 
```
export GYP_MSVS_VERSION=2013
electron-rebuild
```
should work but electron-rebuild is giving me errors so if anyone can test this will be great.

onbjerg:
So the Visual Studio 2013 Redist is not enough? @iiegor

iiegor:
@onbjerg I don't know, I think is not enough... I have Visual Studio 2013, Python 2.7 and the latest node-gyp and works perfect, can you try with this?

onbjerg:
I've tried exporting those variables, but I would not use black-screen if I need to have Visual Studio 2013 installed. Visual Studio bloats my system with a lot of programs that I do not need, because it installs a ton of different packages. I'm never going to use it, so I would rather not.

It is a weird issue, though, as I have no problem building and packaging my own electron projects under Windows, even though some of them also require a rebuild of some modules.

jerch:
@onbjerg "...but I would not use black-screen if I need to have Visual Studio 2013 installed..."

@all
Is there a way to circumvent the node-gyp thingy and deliver binaries for a real distribution package later on? IMHO any unexpericened user would never comply to the "get VS2013 beforehand" issue. As long as it is in dev space - well its normal you would need it ;)

iiegor:
@jerch You mean distribution package for ptyw.js without compiling with node-gyp? If this is what you mean, I think it's possible, try with this pre-compiled binaries by me: https://www.dropbox.com/s/0h8kuo35r508te1/Release.zip?dl=0 *(unzip this in node_modules/ptyw.js/build)*

@onbjerg What are your intentions? I mean, if you want to compile yourself a binary of black-screen on windows I'm afraid you will need to have VS2013 installed or if you can wait, I will compile the respective binary files when black-screen reaches a stable release.

rmobis:
@iiegor When you say `export`, do you mean `SET`? Because my Windows (8.1, tried on both cmd and PowerShell) doesn't seem to have an `export` command/executable and (I'm pretty sure) `SET` is the equivalent of `export` on Linux, so maybe you have an alias?

The command `HOME=~/.electron-gyp node-gyp rebuild --msvs_version="2013" --target=0.30.0 --arch=ia32 --dist-url=https://atom.io/download/atom-shell` also doesn't seem to work on my machine, reporting that `HOME` isn't a command either and I couldn't exactly figure out what you meant with this one.

Also worth noting that for both `electron-rebuild`, `electron` and `electron-package` binaries to be ran, I had to call them from their `node_modules/.bin` folder, as they weren't on my `PATH` by default and maybe won't be for everyone else.

Either way, I followed the steps you described, using the master branch. I managed to get all steps to work (on the second step, `electron-rebuild` worked or at least didn't display any errors) but the final result didn't. I had some issue with `fixedsticky` not being found. While I was writing this I realized I hadn't ran `bower install` and, after doing so and restarting `npm start`, `electron .` worked just fine.

However, I still couldn't get the binary to properly work. At first, I was getting the following error (loosely translated):

```
Uncaught Error: %1 is not a valid Win32 application.
D:\black-screen\.build\Black Screen-win32-ia32\resources\app\node_modules\ptyw.js\build\Release\pty.node", source: ATOM_SHELL_ASAR.js (137)
```

My first thought was that the fact that I was running a x64 version of Node was the problem, so I decided to try the packaging command using `--arch=x64`. It compiled fine and when running it, it seemed to generate a slightly better result (The screen was pretty much fully rendered, I just couldn't type anything), but I still got the an error:

```
Uncaught Error: Invariant Violation: addComponentAsRefTo(...): Only a ReactOwner can have refs. This usually means that you're trying to add a ref to a component that doesn't have an owner (that is, was not created inside of another component's `render` method). Try rendering this component inside of a new top-level component which will hold the ref.", source: D:\black-screen\node_modules\react\lib\invariant.js
```

So, my last attempt was to install a x86 version of Node and rerun all the steps. Unfortunately, that still gave me the same error as the x64 version.

Either way, at least I got the unpackaged version to work.

iiegor:
@rmobis Oh sorry, you are right, it's ``SET`` and regarding the react error (addComponentAsRefTo) it's because the build folder is inside the project folder and the react library is called two times, try moving the build folder outside and try again :ok_hand:

rmobis:
@iiegor Indeed, moving the `.build` folder outside the repo folder does work.

jerch:
@iiegor Just wondering - isn't your instruction listing above to get it build on windows related to git bash or some other unix shell windows env? IMHO `HOME=~/.electron-gyp node-gyp rebuild` won't work on native cmd.exe because of the unixish ENV variable syntax. (can't test it right now)

iiegor:
@jerch Yes, MINGW32. For native shell run ``set HOME=~/.electron-gyp && node-gyp rebuild --msvs_version="2013" --target=0.32.1 --arch=ia32 --dist-url=https://atom.io/download/atom-shell``

onbjerg:
@iiegor I would like to contribute to the project, but I refuse to install Visual Studio 2013. I know this is not a fault in black-screen, but rather node-gyp or Micro$oft's redists. I'll never use VS13 or any VS for that matter, so I find this unnecessary. I've never had this issue before either, when rebuilding npm packages on Windows.

Anyway, I think I'm switching back to Linux in a moment anyway, so this won't be much of an issue for me.

PR #144, opened 2015-09-18
Organize the project tree

iiegor:
### Changes / Improvements
- Project organization
- Use bower for front-end components instead of npm
- Add packaging script for windows
- Icon for windows binary
- Color for branch status
- Fix arrow z-index

**Note:** Please test this on OSX


shockone:
What are the benefits of folders reorganization?

index.html
```diff
@@ -2,25 +2,30 @@
 <html>
 <head>
   <meta charset="utf-8" />
-  <title>Black Screen</title>
-  <link rel="stylesheet" type="text/css" href="compiled/all.css" media="screen"/>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title></title>
+  
+  <!-- TOOD: Merge and minimify -->
+  <link rel="stylesheet" type="text/css" href="dist/main.css" media="screen"/>
```
shockone: We load local files, minification isn't as important in that case.

index.html
```diff
@@ -2,25 +2,30 @@
...
+  <!-- /.TODO -->
+  
+  <script src="dist/scripts/views/main.js"></script>
 </head>
-<body>
-
-<div id="sizes-calculation">1</div>
-
-<div id="black-board"></div>
+<body dir="ltr">
```
shockone: What's `ltr`?

package.json
```diff
@@ -67,7 +65,8 @@
     "preinstall": "npm prune",
     "postinstall": "bower prune; bower install; electron-rebuild",
     "start": "gulp",
-    "package": "electron-packager . 'Black Screen' --overwrite --platform=darwin --arch=x64 --version='0.31.2' --out='/Applications' --icon='./icon.icns'",
+    "package-windows": "electron-packager . \"Black Screen\" --prune --overwrite --platform=\"win32\" --arch=\"ia32\" --version=\"0.33.0\" --out=\"./.build\" --icon=\"./static/icon.ico\"",
```
shockone: That's not exactly how I'd like it to work: we should have a single command no matter what platform a person uses.

static/main.scss
```diff
@@ -270,6 +271,18 @@ body, pre {
...
+        color: #8AD843;
+      }
+
+      &.modified {
+        color: #E9C059;
+      }
+
+      &.deleted {
+        color: #D75857;
+      }
```
shockone: There is a file `colors.scss`. We have to fetch colors from it, because the actual colors will change with themes.

bower.json
```diff
@@ -1,10 +1,12 @@
...
+    "font-awesome": "~4.4.0",
+    "octicons": "~3.1.0",
+    "filament-sticky": "~0.1.6"
   },
   "devDependencies": {
-    "livereload-js": "~2.2.1",
-    "filament-sticky": "~0.1.6"
+    "livereload-js": "~2.2.1"
   }
 }
```
shockone: I'd much rather remove the bower dependencies at all.

index.html
iiegor: Yes, but I thought in a future if we minify the app scripts could reduce a bit the size of the packaged app. Now, this app is on an early stage so I only added a comment, if you want I remove it.

index.html
iiegor: Sorry, this is not necessary I will remove it.

``dir="ltr"`` = Direction of the text: left to right

package.json
iiegor: hmm, ``--platform=all`` could do the trick?

iiegor:
@shockone In my opinion this have several benefits:
- The distribution of the app is more easy across different platforms.
- In a future with this organization is more easier to ignore the files that we don't want in our distribution package.
- The compiled files are located in the corresponding folder name, ``scripts`` and not in ``src``.

There are other benefits but in conclusion this is necessary for simplifying the project files structure and have a more intuitive structure.

RobertoUa:
Renaming folders doesn't seem to be a good idea. 'src' is intuitive, 'static' are not. Why rename compiled to dist? `dist` is distribution which means it should contain nothing but .App (or .exe in case of windows).

iiegor:
@RobertoUa Now, the views folder is outside the src and also the icons (and the future images) so I move all this to a folder called static because this files don't need to be compiled, are static files. The src folder still exists and is for the typescript source files. Also I renamed the ``main`` folder inside src to ``browser`` because the files are related to the browser window and menu.

Regarding the renaming of ``compiled`` to ``dist``... right, it means distribution, so I renamed to dist because the files inside are ready for distribution, they are compiled and these files will be used in the packaged app without any more modification. Anyway if the team feels better using ``compiled`` I can rename it to the original name.

shockone:
Sorry, I won't accept this. There are too many changes at once, and the tree reorganisation is at least questionable.

PR #159, opened 2015-10-14
Adjust build scripts

iiegor:
### Notes
- Arch *ia64* doesn't exist, it's x64.
- Set an exact version for the module **ptyw.js**.
- Create "productName" & "electronVersion" parameters and use them in the different scripts.

In addition, I think it's time to bump the package version to something like 0.0.3-alpha (this last has not been added to the PR yet).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "black-screen",
+  "productName": "Black Screen",
   "version": "0.0.1",
   "description": "A terminal emulator for the 21st century.",
   "main": "src/main/main.js",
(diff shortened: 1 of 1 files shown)


shockone:
Wow, it's awesome we can use variables! Are you sure about `x64`? It does compile for me, and it's described like that here, for example: http://electron.atom.io/docs/v0.28.0/tutorial/using-native-node-modules/

iiegor:
Yes, I think it's correct but it's strange because if you look at the bottom of the docs (http://electron.atom.io/docs/v0.28.0/tutorial/using-native-node-modules/#the-npm-way) they use x64 for building. I googled this and I found: *The IA64 is exclusive to Intel while x64 is used by all.* so x64 is the right option.

shockone:
I totally agree. My concern was whether it'd work.

shockone:
And thank you for the PR!

Issue #141, opened 2015-09-17
Re-locating cursor will retain position for only one action


williara:
After relocation, the cursor will only preserve position for one action (e.g. delete or insert), then repositions to end of line

shockone:
Confirmed.

williara:
Do you have an IRC channel to discuss? I would like to contribute to this project, but it would be nice to get assistance in getting started :)

flawiddsouza:
No IRC but: [![Join the chat at https://gitter.im/shockone/black-screen](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/shockone/black-screen)

PR #167, opened 2015-10-22
Allow cursor to be moved

bmats:
I’m unable to move the cursor away from the end of the line. This issue was reported in #141, but it looks like it was only partially resolved in that ticket.

Changes:
- Move the cursor left and right when the arrow keys are pressed

EDIT: Oops, this PR broke the cursor being positioned by setting the state. I'll work on fixing this.
EDIT 2: Should be fixed now. I updated the changes list above.

### DIFF
--- a/stylesheets/main.scss
+++ b/stylesheets/main.scss
@@ -235,7 +235,7 @@ body, pre {
   // To display even empty rows. The height might need tweaking.
   // TODO: Remove if we always have a fixed buffer width.
   .row {
-    height: $font-size + 4px;
+    min-height: $font-size + 4px;
   }
 }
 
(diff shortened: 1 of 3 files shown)


iiegor:
@bmats I created a PR in your branch to fix a minimal issue.

> When we press the right arrow key we can't move the caret to the last char of the right.

bmats:
Thank you @iiegor, merged.

shockone:
Hey, @bmats. Thank you for the fix. However, with this approach we'd need to manually handle all the other movements  as well, for example `⌘+ ←`, which could be handled by Chrome. Perhaps, we should think of a different way. Maybe reposition the caret only  if the buffer content was changed.

shockone:
Fixed.

PR #169, opened 2015-10-23
Fix getDOMNode deprecation

iiegor:


shockone:
Thank you, somehow I missed it.

Issue #158, opened 2015-10-13
Can't close window with 'exit' or 'ctrl-d'


raboof:
'exit' gives 'command not found', ctrl-d just doesn't do anything.

tcyrus:
```
$ exit
Black Screen: command "exit" not found.
```

iiegor:
This, probably, can be easily fixed by implementing a new built-in for the 'exit' command.

shockone:
Fixed.

PR #170, opened 2015-10-23
Add exit built-in command

bmats:
Call `app.quit()` in a built-in command. This resolves the first half of #158.

"exit" does not show up in autocomplete, even though "clear" does. Should it be added to providers/Executable.ts or is there a better place?


src/Command.ts
```diff
@@ -27,6 +27,10 @@ const executors = {
     },
     clear: (invocation: Invocation, args: Array<string>): void => {
         setTimeout(() => invocation.terminal.clearInvocations(), 0);
+    },
+    exit: (invocation: Invocation, args: Array<string>): void => {
+        var app = require('remote').require('app');
+        app.quit();
```
shockone: It will works correctly now, but we plan to add tabs and split, which would make this behaviour incorrect. Please remove the current terminal instance from this array instead. Perhaps, we might want to exit, if it was the last one. https://github.com/black-screen/black-screen/blob/38a69e28ac1562ed23def97407a0fbce4d5f0619/src/Application.ts#L5

bmats: What is the best way to access `Application` to remove a terminal? Is there a direct way to access the application from within `Command`, or after getting an `Invocation`? Or, would it be better to create an ipc message to handle it in `ApplicationComponent` similar to [this](https://github.com/black-screen/black-screen/blob/075d22e68ca1a4017c8b71b963a049df65423e0e/src/views/ApplicationComponent.ts#L25)?

shockone: I guess we should make `Application` a singleton class accessible from everywhere. We'll have only one instance of it anyway.

iiegor: And what about making an ``application delegate`` class with the common methods.

Something like this:
```ts
class ApplicationDelegate {
  get activeTerminal();
  removeTerminal(index: int);
}
```
Instead of requiring the main application class every time, we have common methods that will send messages through ``ipc`` to an event handler in the ``Application class``.

src/Command.ts
```diff
@@ -27,6 +28,10 @@ const executors: _.Dictionary<(i: Invocation, a: string[]) => void> = {
     },
     clear: (invocation: Invocation, args: string[]): void => {
         setTimeout(() => invocation.terminal.clearInvocations(), 0);
+    },
+    exit: (invocation: Invocation, args: Array<string>): void => {
+        var application = Application.getInstance();
+        application.removeTerminal(application.activeTerminal);
```
shockone: If there are multiple terminals, it doesn't re-render `ApplicationComponent` after removing the terminal.

src/Application.ts
```diff
@@ -4,16 +4,23 @@ import * as _ from 'lodash';
...
+            throw new Error('Use Application.getInstance() instead.');
+        }
+    }
 
-        this.addTerminal();
+    static getInstance(): Application {
+        if (!Application._instance) {
+            Application._instance = new Application();
+        }
+        return Application._instance;
```
shockone: Please use 
```
static get instance()...
```

bmats:
Thank you @shockone, `getInstance()` changed to `instance`.

In order to re-render `ApplicationComponent` when the terminals change, I modified `Application` to emit events when it changes (similar to `Terminal`) and subscribed `ApplicationComponent` to these. This is in https://github.com/bmats/black-screen/tree/exit-command-app-emitter. Is this approach OK, or is there a better solution?

shockone:
@bmats, I'm sorry it took me so long. Could you please update it, because it's conflicts now?

bmats:
No problem, @shockone. I fixed the conflicts.

shockone:
Thank you, @bmats.

PR #181, opened 2015-11-25
tiny optimization to make the ol' git clone and cd work for bash and zsh

pavgup:
Almost not worth it, but you know, why not.  I tested a quick copy paste of the updated command and I think it's comfy in two universes now.


shockone:
Thank you.

PR #184, opened 2015-12-08
Integrated iterm2-colors

kevin-smets:
Added iterm2-colors as an npm dependency. With this package, you get all iterm2 color schemes for free: (https://github.com/mbadolato/iTerm2-Color-Schemes).

To fully leverage this, there should be some preference pane where you can toggle what CSS gets loaded. For now, you can easily toggle it in source.

E.g. Duotone Dark theme:
screen shot 2015-12-08 at 21 46 06

E.g. Darkside theme:
screen shot 2015-12-08 at 21 47 04

E.g. Solarized Dark theme:

screen shot 2015-12-08 at 22 02 47

It does not work with all themes perfectly, since black-screen does not use all the same variable names. Maybe a good thing to do is to align to the iTerm2 color schemes (since it matches so close already)?


shockone:
Hi, @kevin-smets. I didn't realize it would be so easy to add themes support. I totally agree that we should align the color names. However, I can't accept this feature in such a state. We need those two things you mentioned to merge it:

* Match the colors
* Have the ability to change a theme without modifying the code. 

About the second one, we could add a top-level class like `solarize-theme`, and then depending on it would change the actual colors. For now the theme could depend on a global variable, for example. 

I will do that eventually, but if you'd like it to be shipped faster, you're welcome to finish the work. :)

shockone:
As this issue suggests, we could have a dotfile for configuration.

shockone:
Hi, @kevin-smets. I've made a decision that Black Screen won't support themes. Because I'm afraid people will make a white one. Just kidding. Seriously speaking, I'm afraid they won't be of good quality.

kevin-smets:
No problemo, it's your project :D. Sorry that I left this PR open.

If quality is a concern, you could certainly limit to only the quality themes, or a personal selection. But this collection is an existing one which serves a lot of iterm2 users. I do see it as a missed opportunity to easily integrate a load of already existing and battle tested themes.

Is this out of scope forever? Or maybe I could convince you with a more integrated solution?

laughedelic:
It's sad to see that such a nice project is missing an opportunity to be more customisable. I think that if the quality of the themes is a real concern, then shipping only a selected set of themes would be a solution. But just closing a pull-request (it's already done and @kevin-smets is open for improvements!) seems to me a strange decision.. "(

crutchcorn:
Albeit just a user, I figure I'll drop my 2 cents here. While it is @shockone's project and whatever he says should be law, it would be NICE to see the customisability already worked on to be implemented here. Especially with this much support already available (referring to the iTerm2 color repo). I honestly have to say that I agree with @laughedelic with his thoughts on this. I think that a user should have freedom to make his/her set up as hideous as possible - assuming that's what they want to do. Freedom, my good sir, freedom. ;)

shockone:
It is possible indeed, @crutchcorn. All you need is to change the colors in the source code and rebuild the project. 

If we have many requests for themes, I might implement it, but for now it sounds unimportant comparing to other things at hand.

crutchcorn:
Still though, while I don´t particularly mind doing work with the source and rebuilding - it would be less work to just have it as an option.

Ooooh, I COMPLETLEY understand that!!! I thought you were just opposed to the idea as a whole and will never implement it. I would rather have a release for the project than just having some silly config stuff working without the functionality or stability/usability there.

PR #185, opened 2015-12-10
Add fixPath() to Main.ts, allows correct $PATH on OSX

### LINKED ISSUE #50: The packaged version doesn't read PATH
Extracting @alebcay's comment as a separate issue.

I'm also getting some other weird stuff with `ls` coloring and formatting. This is what I get when I run it through `gulp` - everything works the way it should:

q4ysyk9

but then, when I do `npm run package` and run it from the standalone app bundle, I get this:

af2jiml

It seems that the `gulp` version is reading my path - `which ls` returns my `ls` that I installed from MacPorts. However, the standalone doesn't seem to do that, and is instead using `/bin/ls`. Even so, I still can't figure out why it's only outputting the first line out output with `/bin/ls` (it only shows "Applications", nothing else).

moollaza:
Fixes #50 

I tested this locally and it seems to work. 

I've never used TypeScript before, so I hope this is the correct approach. Please let me know if it isn't!

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
   ],
   "electronVersion": "0.35.0",
   "dependencies": {
+    "fix-path": "1.1.0",
     "fixed-sticky": "^0.1.6",
     "font-awesome": "4.4.0",
     "fuzzaldrin": "2.1.0",
--- a/src/main/Main.ts
+++ b/src/main/Main.ts
@@ -2,8 +2,12 @@ const app = require('app');
 const BrowserWindow = require('browser-window');
 const IPC = require('electron').ipcMain;
 let menu = require('./Menu');
+let fixPath = require('fix-path');
 
 let browserWindow: any = null;
+
+// Fix the $PATH on OS X
+fixPath();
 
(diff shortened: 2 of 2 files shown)


jdsimcoe:
@moollaza I built your files and when I try to run `atom .` it gives me this error:

```
LSOpenURLsWithRole() failed for the application /Applications/Atom.app with error -10810.
```

cpdean:
@jdsimcoe are you running `atom .` from within a terminal inside tmux?

moollaza:
@jdsimcoe I tested with `npm start` vs `npm run package` -- the latter, compiled version had the same `$PATH` as the former.

When I run `atom .` nothing happens on the current master branch of black-screen for me...

jdsimcoe:
@cpdean No I'm running `atom .` from within `black-screen`.

moollaza:
@jdsimcoe I'm able to replicate now. It looks like this is an Atom specific bug: https://github.com/atom/atom/issues/3093 -- I see you commented over there :+1: 

Other programs seem to work fine, for example `subl .`

shockone:
Thank you, @moollaza.

moollaza:
@shockone no problem! Thanks for merging :smiley:

PR #188, opened 2015-12-14
Fix spelling on SuggestionCompoonent

j-allard:
While I was hacking away on adding clickable suggestions, I noticed the misspelling in `SuggestionCompoonent`.
Is there any reason why it was named that way ? 
Anyway, this PR renames the React component `SuggestionCompoonent` to `SuggestionComponent`
I've run ag on the codebase to make sure I caught all occurences. Looks like the component is not used outside of this single file.


shockone:
@j-allard, thank you. I try to have as few rules about PRs as possible because people hate rules. So, in case you plan to contribute further, you don't have to create a separate PR for this, just fix what you think is wrong along with some feature. But separate PRs are good too. :)

PR #189, opened 2015-12-15
Add clickable suggestions for AutocompleteComponent

j-allard:
This does three main things:
1) Hovering a suggestion changes `highlightedSuggestionIndex`.
2) A click on a suggestion calls `selectAutocomplete()` (i.e it autocompletes the input with the clicked suggestion)
3) `suggestionComponent` is now styled with `{cursor: "pointer"}` so it actually looks clickable.

Any thoughts ? :-)


shockone:
You are awesome, @j-allard! I planned to do that but never had time.

PR #197, opened 2016-01-05
Replace deprecated restart() method with reload()

j-allard:
This replace the deprecated `restart()` method with `reload()`.
From what I've seen, they should both behave the same way.
Fix https://github.com/shockone/black-screen/issues/196


shockone:
Thank you, @j-allard.

PR #201, opened 2016-01-14
Fix TODO link

mortonfox:
The relative "issues/58" link pointed within the repo, so that didn't work.


shockone:
Thank you, @mortonfox.

PR #238, opened 2016-02-08
Fix spelling of "doesn't".

marnen:


shockone:
Thank you, @marnen.

marnen:
You're most welcome!

PR #264, opened 2016-03-13
Removed redundant spaces in commands before adding to session history

mkenne11:
This change removes leading, trailing and multiple spaces from commands before they are added to session history.


shockone:
Thank you, @mkenne11.

PR #269, opened 2016-03-19
Filtered out suggestion items whose score is zero

mkenne11:
Removed items with a (fuzzy) score of zero from the autocomplete suggestions list.

This change is to address the FIXME comment in https://github.com/shockone/black-screen/blob/master/src/Autocompletion.ts#l15

### DIFF
--- a/src/Autocompletion.ts
+++ b/src/Autocompletion.ts
@@ -15,9 +15,12 @@ export default class Autocompletion implements i.AutocompletionProvider {
         return Promise.all(providers.map(provider => provider.getSuggestions(job))).then(results =>
             _._(results)
                 .flatten()
+                // .forEach((suggestion: Suggestion) =>
+                //         suggestion.score = score(suggestion.value, suggestion.getPrefix(job))
+                // )
                 .filter((suggestion: Suggestion) =>
-                    !suggestion.shouldIgnore(job) &&
-                    score(suggestion.value, suggestion.getPrefix(job)) > 0)
+                        !suggestion.shouldIgnore(job) &&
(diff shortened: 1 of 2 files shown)


src/Autocompletion.ts
```diff
@@ -12,11 +12,12 @@ export default class Autocompletion implements i.AutocompletionProvider {
...
         let providers = specializedProviders.length ? specializedProviders : PluginManager.genericAutocompletionProviders;
 
-        // FIXME: skip suggestions that have 0 score.
         return Promise.all(providers.map(provider => provider.getSuggestions(job))).then(results =>
             _._(results)
                 .flatten()
-                .filter((suggestion: Suggestion) => !suggestion.shouldIgnore(job))
+                .filter((suggestion: Suggestion) =>
+                    !suggestion.shouldIgnore(job) &&
+                    score(suggestion.value, suggestion.getPrefix(job)) > 0)
```
shockone: I should have made the comment clearer, but the thing is, scoring is quite time consuming, and we'd like to avoid doing it twice.

mkenne11:
Thanks for the feedback :) Ill work on a method that only calls score once when modifying the suggestions list.

shockone:
@mkenne11, do you still plan to finish this up? I can do that, but I don't want to steal your contribution. :)

mkenne11:
Hey @shockone go ahead and finish please. It turned out to be complicated than I first though. Am interested to see how you will address it.

shockone:
Here it is. https://github.com/shockone/black-screen/commit/ff8c878f682ae2acac5fb3590559a91e6599eda7

Issue #280, opened 2016-04-10
Switching between tabs removes command history


mkenne11:
If the user switches from one tab (after running commands) to a second tab, then back to the first tab - (in the first tab) the results of commands run are shown however commands run are no longer displayed.

As an example see the screenshots below. In this example the following scenario occured:
- Initially 3 tabs were opened.
- Three commands were run in tab # 3. See Screenshot 1.
- The user switches to tab # 2 then back to tab # 3.
- In tab # 3 the results of the 3 commands are shown however the individual commands are no longer shown. See Screenshot 2.

Screenshot 1.
![session history - before switch](https://cloud.githubusercontent.com/assets/12181805/14410133/41f96de2-ff6a-11e5-8294-d8e8853b7432.PNG)

Screenshot 2.
![session history - after switch](https://cloud.githubusercontent.com/assets/12181805/14410135/46f744ae-ff6a-11e5-87f2-71323278998a.PNG)

shockone:
There is more to read about this bug in the linked pull request.

mkenne11:
Thanks @shockone - yep I saw your comments in the pr. I'll have an attempt at trying to implement the suggested fix hopefully I'll find time. 

BTW. I'm keen to learn typescript and react and this seems like a good project to start with :)

PR #281, opened 2016-04-10
Tab command history fix

mkenne11:
This change addresses issue #280 where individual commands are no longer displayed after switching tabs.

### DIFF
--- a/src/views/1_ApplicationComponent.tsx
+++ b/src/views/1_ApplicationComponent.tsx
@@ -72,7 +72,7 @@ export default class ApplicationComponent extends React.Component<{}, State> {
             }
         }
 
-        if (event.metaKey && event.keyCode === KeyCode.T) {
+        if ((event.metaKey || event.ctrlKey) && event.keyCode === KeyCode.T) {
             if (this.tabs.length < 9) {
                 this.addTab();
                 this.setState({sessions: this.activeTab.sessions});
--- a/src/views/4_PromptComponent.tsx
+++ b/src/views/4_PromptComponent.tsx
@@ -163,6 +163,11 @@ export default class PromptComponent extends React.Component<Props, State> imple
             scrollToTop = <a className="scroll-to-top" onClick={this.handleScrollToTop.bind(this)}><i className="fa fa-long-arrow-up"/></a>;
         }
 
+        let contentEditable: boolean = this.props.status === e.Status.NotStarted ||
+             this.props.status === e.Status.InProgress;
+        let promptValue: String;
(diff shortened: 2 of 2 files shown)


mkenne11:
Btw. I'm sure if this PR is a good fix for the issue #280 I identified.

@shockone this PR was building successfully previously on CircleCI. I rebased the updated "master" branch onto it before re-pushing the branch to my forked repo. There appears to be a clash between my local machine (Ubuntu) vs. the current build machine Mac OS/Darwin.

mkenne11:
I see other recent PRs have failed too with the same error:
```
npm ERR! Linux 3.13.0-83-generic
npm ERR! argv "/home/ubuntu/nvm/versions/node/v5.1.0/bin/node" "/home/ubuntu/nvm/versions/node/v5.1.0/bin/npm" "install"
npm ERR! node v5.1.0
npm ERR! npm  v2.13.5
npm ERR! code EBADPLATFORM

npm ERR! notsup Unsupported
npm ERR! notsup Not compatible with your operating system or architecture: fsevents@1.0.8
npm ERR! notsup Valid OS:    darwin
npm ERR! notsup Valid Arch:  any
npm ERR! notsup Actual OS:   linux
npm ERR! notsup Actual Arch: x64
```
(Sorry I should have done some more investigation first before raising my previous comment).

I did some investigation and saw users reporting that selecting specific versions of nodejs/npm downgraded the error to a warning:
https://github.com/npm/npm/issues/10768

I also saw a mention of the error with optional dependencies when using shrinkwrap:
https://github.com/npm/npm/issues/2679

Not sure if those links help.

shockone:
Hey, @mkenne11. It was my fault that the tests were red. Now they're fixed in master.

However, I'm afraid I'll have to decline these changes. I don't remember the details, but setting the prompt value in the render method caused some troubles (perhaps because of lifecycle callbacks), so I isolated it to the method `setDOMValueProgrammatically`.

Furthermore, the root of this issue is that I took a shortcut and modeled tabs incorrectly: instead of having a list of jobs inside every tab component, we have a single list of jobs, which re-renders based on the active tab. This re-rendering is exactly what causes the browser to lose data in inputs. 

So, the fix should be to move `SessionComponent`s from `ApplicationComponent` into `TabComponent`. I suspect this fix will also simplify the way we destroy things, but I'm not sure about that.

Despite that, I want to thank you for investing your time into Black Screen.

mkenne11:
No worries @shockone :) in hindsight my suggest fix seems a hack - it addresses the issues symptoms not the root issue.

PR #299, opened 2016-05-01
Tab close buttons

mkenne11:
This change adds close buttons to tabs - these are visible when more than one tab is open. A sample of the close buttons is shown in the following screenshot.

image

The close button changes on focus. See additional sample screenshot below.

image

### DIFF
--- a/src/views/1_ApplicationComponent.tsx
+++ b/src/views/1_ApplicationComponent.tsx
@@ -116,7 +116,8 @@ export default class ApplicationComponent extends React.Component<{}, State> {
                               activate={() => {
                                 this.activeTabIndex = index;
                                 this.setState({ sessions: this.activeTab.sessions });
-                              }}>
+                              }}
+                              closeHandler={this.closeSelectedTab.bind(this)}>
                 </TabComponent>
             );
         }
@@ -128,7 +129,7 @@ export default class ApplicationComponent extends React.Component<{}, State> {
                               activate={() => {
                                    this.activeTab.activateSession(session);
                                    this.setState({ sessions: this.activeTab.sessions });
-                               }}>
+                              }}>
             </SessionComponent>
         );
 
@@ -138,6 +139,14 @@ export default class ApplicationComponent extends React.Component<{}, State> {
                 <div className="active-tab-content">{sessions}</div>
             </div>
         );
+    }
+
+    closeSelectedTab(event: any) {
+        let selectedTabIndex: number = Number(event.target.value);
+        this.closeTab(this.tabs[selectedTabIndex]);
+        this.setState({sessions: this.activeTab.sessions});
+        event.stopPropagation();
+        event.preventDefault();
     }
 
     public closeSession(sessionToClose: Session) {
--- a/src/views/TabComponent.tsx
+++ b/src/views/TabComponent.tsx
@@ -8,10 +8,12 @@ export interface TabProps {
     isActive: boolean;
     activate: () => void;
     position: number;
+    closeHandler: () => void;
 }
 
-export const TabComponent = ({ isActive, activate, position }: TabProps) =>
+export const TabComponent = ({ isActive, activate, position, closeHandler }: TabProps) =>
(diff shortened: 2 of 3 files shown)


mkenne11:
@shockone I am not sure what you had in mind when it came to mouse/ui design for tabs. Please let me if this is close to something you expect or not. I am happy to make changes to get closer to what you wanted.

src/views/1_ApplicationComponent.tsx
```diff
@@ -72,7 +72,7 @@ export default class ApplicationComponent extends React.Component<{}, State> {
             }
         }
 
-        if (event.metaKey && event.keyCode === KeyCode.T) {
+        if ((event.metaKey || event.ctrlKey) && event.keyCode === KeyCode.T) {
```
shockone: This is not related to the subject. Please remove this change.

src/views/TabComponent.tsx
```diff
@@ -8,12 +8,14 @@ export interface TabProps {
     isActive: boolean;
     activate: () => void;
     position: number;
+    closetab: () => void;
```
shockone: Should be `closeHandler`.

src/views/1_ApplicationComponent.tsx
```diff
@@ -140,6 +141,20 @@ export default class ApplicationComponent extends React.Component<{}, State> {
...
+        let selectedTabIndex: number = Number(event.target.value);
+
+        if (selectedTabIndex === this.activeTabIndex) {
+            this.closeSession(this.activeTab.activeSession());
+        } else {
+            this.closeTab(this.tabs[selectedTabIndex]);
+            if (selectedTabIndex < this.activeTabIndex)
+                this.activeTabIndex -= 1;
+            this.setState({sessions: this.activeTab.sessions});
+        }
```
shockone: It really should be the same as `event.metaKey && event.keyCode === KeyCode.W` handling (except selecting the  correct tab to close).

shockone:
Thank you for the contribution, @mkenne11. Design-wise I'd like the close button to appear on the left side (as in Safari tabs), and also it would be cool it it appeared only on hover to have less visual noise.

mkenne11:
@shockone I believe I made the changes you requested. Please let me know if any extra changes are required :)

shockone:
Thanks, @mkenne11. I'll merge it. I think I know a better way to handle closing, but I can modify it by myself instead of bothering you. Also, it looks buggy that the tab number is not aligned properly, but we can fix that separately. 

Two more things:
* Please don't use color values, try to get them from the map. You'll find a lot of examples in the CSS file. It will help when we modify the theme or implement themes support.
* How strong do you feel about having the top line? I thought that distinguishing tabs by color should be enough. Besides, it makes the number move down, which also looks buggy.

mkenne11:
No worries @shockone. I can see your point about using colour values from the map - you have a standard colour palette and you'd like like to use that - definitely makes things more consistent.

I don't feel that strongly about having the top border, it was an idea I had but I'm not wedded to it. And can definetly see your point that moving the number down makes it look buggy.

shockone:
@mkenne11, I've eliminated the need for the `value` attribute by binding the index to the handler as we iterate over them. https://github.com/shockone/black-screen/commit/fcd7184714e95eae1170b5286a5489fdd4166a39

mkenne11:
Thanks @shockone that looks cleaner.

shockone:
I've messed with CSS quite a bit. Now it looks closer to the original tabs design. https://github.com/shockone/black-screen/commit/23921e449266d942a146aae628fa1cf9c7569f40

On hover:

On icon hover:


Hopefully I haven't offended you by modifying your work.

mkenne11:
No probs at all I'm not offended :) your the main author of the project and have a vision for how things should look and feel. Plus I'd expect you'd want the "cleanest" commits possible to minimise code maintenance over time.

shockone:
Awesome, @mkenne11. I'm glad you're cool with it.

PR #306, opened 2016-05-07
Added new tab shortcut ctrlKey+T

mkenne11:
Added new tab shortcut Ctrl+T for users with PC-based keyboards. This compliments the Command+T shortcut for Mac-based keyboards.

### DIFF
--- a/src/views/1_ApplicationComponent.tsx
+++ b/src/views/1_ApplicationComponent.tsx
@@ -72,7 +72,7 @@ export default class ApplicationComponent extends React.Component<{}, State> {
             }
         }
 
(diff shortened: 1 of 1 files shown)


shockone:
Sorry, but I can't accept this pull request.

It is not a scalable solution. I understand the need for different key bindings, but I imagined it differently. We'd have an object representing mappings from actions to keys. This particular piece of code would look like `if (keyBindings.isNewTab(event)) {`. But it requires much more efforts. We can create an issue and discuss implementation details.

mkenne11:
No worries @shockone - I'll create a new issue to address this.

PR #349, opened 2016-05-22
Fix wrong font in output cut messages

Sumolari:
Extending `.octicon` in `.output-cut` made non-icon text to not inherit proper font, using system's default.
With this fix `.output-cut:before` extends `.output-cut` so icon is properly displayed but the rest of the text inherits proper font as expected.


shockone:
Already fixed in master. But thank you, @Sumolari.

Issue #380, opened 2016-05-25
Dragging a file onto program causes file to take over display.


tanishq-dubey:
This is the result of dragging the `LICENSE` file and dropping it on the window.

![screen shot 2016-05-24 at 10 58 40 pm](https://cloud.githubusercontent.com/assets/3333201/15529596/4a8fbf0e-2203-11e6-9d32-bb5dcf9c538f.png)

Seems like the dropped file is replacing the terminal (just as a webpage or document would)

tanishq-dubey:
Fixed in PR #381

PR #381, opened 2016-05-25
Drag and drop fix

tanishq-dubey:
Fixed issue (#380) where dragging a file onto terminal window caused file to be opened, getting rid of the terminal.

### DIFF
--- a/src/views/4_PromptComponent.tsx
+++ b/src/views/4_PromptComponent.tsx
@@ -348,8 +348,8 @@ export class PromptComponent extends React.Component<Props, State> implements Ke
     }
 
     private handleDrop(event: DragEvent) {
-        console.log(event.dataTransfer.files[0].path)
-        this.prompt.setValue(this.prompt.value + (event.dataTransfer.files[0].path))
+        console.log(event.dataTransfer.files[0].path);
+        this.prompt.setValue(this.prompt.value + (event.dataTransfer.files[0].path));
         this.setDOMValueProgrammatically(this.prompt.value);
     }
 }
--- a/src/views/4_PromptComponent.tsx
+++ b/src/views/4_PromptComponent.tsx
@@ -348,7 +348,6 @@ export class PromptComponent extends React.Component<Props, State> implements Ke
     }
 
     private handleDrop(event: DragEvent) {
-        console.log(event.dataTransfer.files[0].path);
         this.prompt.setValue(this.prompt.value + (event.dataTransfer.files[0].path));
         this.setDOMValueProgrammatically(this.prompt.value);
     }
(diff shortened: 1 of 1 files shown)


shockone:
It's not the expected behaviour, @tanishq-dubey, sorry. It should put the path to the dropped file into the prompt instead.

tanishq-dubey:
Gotcha, I can have a fix for that in a couple days if you wish. This code is already reporting the absolute path of the file dropped in the console, so its just a matter of filling the prompt with the same path.

shockone:
I can't assign you, @tanishq-dubey, but I'd be glad if you did it. :)

tanishq-dubey:
On it!

tanishq-dubey:
Drag and drop should be fixed!

src/views/4_PromptComponent.tsx
```diff
@@ -144,11 +162,11 @@ export default class PromptComponent extends React.Component<Props, State> imple
...
-                                                  caretOffset={$(this.commandNode).caret("offset")}
-                                                  onSuggestionHover={this.highlightSuggestion.bind(this)}
-                                                  onSuggestionClick={this.applySuggestion.bind(this)}
-                                                  highlightedIndex={this.state.highlightedSuggestionIndex}
-                                                  ref="autocomplete"/>;
+                caretOffset={$(this.commandNode).caret("offset")}
+                onSuggestionHover={this.highlightSuggestion.bind(this)}
+                onSuggestionClick={this.applySuggestion.bind(this)}
+                highlightedIndex={this.state.highlightedSuggestionIndex}
+                ref="autocomplete"/>;
```
shockone: Come on, this shouldn't have changed.

shockone:
So, as I understand, it only works if you drop directly on a prompt. If I drop on any other region, nothing happens. Is that correct?

tanishq-dubey:
Fixed that formatting issue you referenced, sorry about that, my sublime setup must have done something. To answer your other question, yes, the drop only works if you drop onto the prompt. I was getting quite a few errors if I tried to have a document wide drop policy both from trying to integrate it into `4_PromptComponent` or `3_JobComponent`

src/views/4_PromptComponent.tsx
```diff
@@ -327,4 +346,10 @@ export class PromptComponent extends React.Component<Props, State> implements Ke
             stopBubblingUp(event);
         }
     }
+
+    private handleDrop(event: DragEvent) {
+        console.log(event.dataTransfer.files[0].path);
```
shockone: A forgotten log.

shockone:
Thank you, @tanishq-dubey.

PR #391, opened 2016-05-25
Add missing comma causing error in npm install.

carolineartz:


shockone:
Thank you , but I've noticed this PR too late.

PR #434, opened 2016-06-01
Update README.md

mldangelo:
Added a line explaining that brew should be updated if the black-screen cask is not found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,13 @@ We aim to be compatible at least with [VT100](https://en.wikipedia.org/wiki/VT10
 Install
 ------------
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,10 +25,14 @@ We aim to be compatible at least with [VT100](https://en.wikipedia.org/wiki/VT10
 Install
 ------------
 
+black-screen can be installed with [homebrew](http://brew.sh/). 
+
 ```bash
 brew cask install black-screen
```
shockone: Just adding `brew update` before this line should be enough.

Issue #454, opened 2016-06-03
Cannot cd (change directory) to directory with name that contains spaces


adchilds:
Using the latest code on master, I cannot cd to a directory that contains spaces in it's name.

For example, if my current working directory is /Users/adchilds/Desktop/ and I have a directory inside my cwd named "Desktop Items" I receive an error saying this directory cannot be found. See the following screenshot:

<img width="1462" alt="cd_directory_with_space" src="https://cloud.githubusercontent.com/assets/1317452/15767081/d66a24fc-290a-11e6-99a5-100d9839a032.png">

This looks to be caused by the fact that when we parse the path in the Common#resolveFile function, we're preserving escape characters such as \ in the path when we really should be getting rid of them.

I have a fix and want to submit a pull-request for this but would appreciate some scrutiny since I'm not entirely familiar with how this may affect the application as a whole.

Thank you!

shockone:
Fixed.

PR #455, opened 2016-06-03
Issue #454: Fix issue with directory not being found during cd (chang…

adchilds:
…e directory) if path contained spaces.

This fixes the issue noted in https://github.com/shockone/black-screen/issues/454 though I'm not sure how updating this function may affect the application as a whole, so input is needed.

### DIFF
--- a/src/utils/Common.ts
+++ b/src/utils/Common.ts
@@ -181,7 +181,22 @@ export function resolveDirectory(pwd: string, directory: string): string {
 }
 
 export function resolveFile(pwd: string, file: string): string {
-    return Path.resolve(pwd, file.replace(/^~/, homeDirectory));
+    let parsedPath = sanitizeSpecialCharacters(file);
+
+    return Path.resolve(pwd, parsedPath);
+}
+
+/**
+ * Replace any special characters to create a fully qualified path.
+ *
+ * '~' - the user's home directory
+ * '\' - escapes space character in file/directory path
+ */
+export function sanitizeSpecialCharacters(path: string): string {
+    let parsedPath = path.replace(/^~/, homeDirectory);
+    parsedPath = parsedPath.replace(/\\/, "");
+
+    return parsedPath;
 }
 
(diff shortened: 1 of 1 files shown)


src/utils/Common.ts
```diff
@@ -181,7 +181,22 @@ export function resolveDirectory(pwd: string, directory: string): string {
...
+    return Path.resolve(pwd, parsedPath);
+}
+
+/**
+ * Replace any special characters to create a fully qualified path.
+ *
+ * '~' - the user's home directory
+ * '\' - escapes space character in file/directory path
+ */
+export function sanitizeSpecialCharacters(path: string): string {
```
adchilds: This function technically doesn't need to be exported. Let me know if that's an issue.

shockone:
Hey, @adchilds. This is a good contribution! But I'm afraid this place is too late to fix this issue. Stripping special characters is a job of a lexer. If you'd like to redo the PR, you'd need a few things:

* Change the `lex` function to throw away those backslashes https://github.com/shockone/black-screen/blob/master/src/CommandExpander.ts#L28-L28
* Change a test to reflect that https://github.com/shockone/black-screen/blob/master/test/lexer_spec.ts#L22-L22
* (This one is a bit harder) Since we need to pass the exact same input to a shell - otherwise it might have a different meaning - and we changed the lexed input, we can't rely on it anymore. So, we'd want to modify the constructor https://github.com/shockone/black-screen/blob/master/src/PTY.ts#L14-L14 to take a single string parameter instead of the first two, and pass it `job.prompt.value`. There are a few instantiations, which would need to be changed to.

Please let me know if it makes sense and if you're willing to try doing it.

adchilds:
Sure, I'll take a look at that within the next day. Thanks for the input!

shockone:
I just realized that the third step wouldn't work: then it wouldn't expand aliases, history substitution and things like that. Sorry about that, I hope you haven't started your work yet. I think I'll fix it by myself.

PR #466, opened 2016-06-06
Tab name functionality

mkenne11:
Added functionality to edit tab  #name. Where tab names are longer than the field width the name is truncated and a tooltip shown.
Note, the user can only edit the tab name on the active-tab.

### DIFF
--- a/src/views/TabComponent.tsx
+++ b/src/views/TabComponent.tsx
@@ -16,7 +16,8 @@ export interface TabProps {
 export enum TabHoverState {
     Nothing,
     Tab,
-    Close
+    Close,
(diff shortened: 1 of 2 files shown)


shockone:
I'm sorry, but I don't think this feature is a good fit for Black Screen. I try to focus on simplicity, and it's more suited for "configure everything" Linux-style kind of programs.

mkenne11:
Not a problem @shockone. It was a good learning exercise for me about the use of react components :) 

If you have time, I would be keen to hear your ideas about adding a tab name feature - if you feel it's worth adding at all.

PR #495, opened 2016-06-10
Show node version on README.md

danderu:
I think it would be useful to show that this project only works on node versions >= 6.0.0, since there are some issues of users asking why it's not working below that one. i.e. shockone#492


shockone:
I don't think it'll help. The proper fix would be to check user's node version in pre- script (e.g. preinstall) and display a meaningful error if it's lower than expected. But it's not trivial to compare versions without help of some library because `10.0` should be greater than `6.0`.

danderu:
Good idea, I'll do that, then

PR #498, opened 2016-06-10
Fix issue with relative path

denieler:
If file is existing it is trying to execute the command using `ShellExecutionStrategy`, but if not we don't pass this command to any Strategy. My proposition is to check is this command has correct path format, and if so - execute it by `ShellExecutionStrategy`. If not - just go to the next available Strategy.

### DIFF
--- a/src/CommandExecutor.ts
+++ b/src/CommandExecutor.ts
@@ -40,7 +40,8 @@ class ShellExecutionStrategy extends CommandExecutionStrategy {
     static async canExecute(job: Job) {
         return loginShell.preCommandModifiers.includes(job.prompt.commandName) ||
             await this.isExecutableFromPath(job) ||
-            await this.isPathOfExecutable(job);
+            await this.isPathOfExecutable(job) ||
+            this.isValidPath(job.prompt.commandName);
     }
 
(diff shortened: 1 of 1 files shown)


shockone:
What's the use case, @denieler? I'm not sure why we'd try to invoke a non-existing file.

shockone:
Oh, I've read your comment in the issue. But we should fix the bug that causes it, namely the answer to the questions `isPathOfExecutable` is currently incorrect.

denieler:
@shockone the bug is that if the file is not existing - we skip current executor and show that black-screen do not have such command. But native behavior is to show bash error, that file is not existing. That's why I propose to check if the command is a path (filename, relative path, absolute path, etc) - then execute it with shell (bash), but if not - skip this shell strategy

shockone:
@denieler got it, the idea is to show an appropriate error message. I'll delay merging for now, if you don't mind: I plan to write a scanner that would distinguish paths from other kinds of tokens, and we'll be able to implement this feature much easier.

denieler:
@shockone sure! thank you ) just show your solution later ) it's interesting)

PR #499, opened 2016-06-10
Added preinstall check for Node.JS version

danderu:
After the suggestion in my previous PR, I'm adding a preinstall check to show a meaningful error when a user tries to install black-screen on an unsupported version of node.js or npm.


shockone:
Good job, @danderu. A few things:
* Please merge the two files and move them into the `housekeeping` directory.
* It might be easier to read node version from `.nvmrc`.
* I'd really prefer it to run twice as slow but without an extra dependency.
* `semver` seems to be available by accident: it's a transient dependency. Please explicitly add it to `devDependencies`.

PR #517, opened 2016-06-17
Toggle to last tab

ekaragodin:


ekaragodin:
In popular programs with tabs (google chrome, iTerm2, ...) I always have ability to toggle to the last tab.

src/views/1_ApplicationComponent.tsx
```diff
@@ -107,7 +107,11 @@ export class ApplicationComponent extends React.Component<{}, State> {
         }
 
         if (event.metaKey && event.keyCode >= KeyCode.One && event.keyCode <= KeyCode.Nine) {
-            const newTabIndex = parseInt(event.key, 10) - 1;
+            let newTabIndex = parseInt(event.key, 10) - 1;
+
+            if (newTabIndex === 8) {
+                newTabIndex = this.tabs.length - 1;
+            }
```
shockone: Wow, I didn't know about this feature. Can we do `const newTabIndex = (event.keyCode === KeyCode.Nine  this.tabs.length : parseInt(event.key, 10)) - 1;` instead?

shockone:
Thank you. Good job!

PR #537, opened 2016-06-27
Fixing Common.stat method to accept a broken symlink.

### LINKED ISSUE #533: Invalid symlinks are breaking autompletion
As soon as there is a invalid symlink in `home` folder or in current folder, the autocompletion doesn't work.

Can be simulated on master (v 0.1.5).

Way to reproduce:
in your home directory run `ln -s non-exiting-file broken-symlink`
tada, the whole autocompletion is broken.

Logs:
Very mysterious message comming from compiled `PromptComponent`:
/home/vicek/workspace/black-screen/compiled/src/views/4_PromptComponent.js:5 Uncaught (in promise) Error: ENOENT: no such file or directory, stat '/home/vicek/broken-symlink'
    at Error (native)

Way to fix:
In https://github.com/shockone/black-screen/blob/87b577b5692867bebbafac483473dd5e292239a4/src/utils/Common.ts#L63 either check if file exists or not reject on error (it's not resilient to drop the whole autocompletion because of one dodgy file)

viktomas:
- Using lstat method (which is not following symlinks) instead of stat.
- Adding mock-fs test dependency to enable unit test

fixes #533

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,11 +44,13 @@
   "devDependencies": {
     "@types/chai": "3.4.22-alpha",
     "@types/mocha": "2.2.21-alpha",
+    "@types/mock-fs": "3.6.22-alpha",
     "chai": "3.5.0",
     "devtron": "1.2.1",
     "electron-builder": "5.7.0",
     "electron-prebuilt": "1.2.5",
     "mocha": "2.5.3",
+    "mock-fs": "3.9.0",
     "npm-check-updates": "2.6.7",
     "spectron": "3.2.3",
     "ts-node": "0.9.3",
--- a/src/utils/Common.ts
+++ b/src/utils/Common.ts
@@ -60,7 +60,7 @@ export function recursiveFilesIn(directoryPath: string): Promise<string[]> {
 
 export function stat(filePath: string): Promise<fs.Stats> {
     return new Promise((resolve, reject) => {
-        fs.stat(filePath, (error: NodeJS.ErrnoException, pathStat: fs.Stats) => {
+        fs.lstat(filePath, (error: NodeJS.ErrnoException, pathStat: fs.Stats) => {
             if (error) {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -44,11 +44,13 @@
   "devDependencies": {
     "@types/chai": "3.4.22-alpha",
     "@types/mocha": "2.2.21-alpha",
+    "@types/mock-fs": "3.6.22-alpha",
```
viktomas: I was wondering if the type definitions need to have the same version as the main module but I took example in the rest. (I'm new to TypeScript)

src/utils/Common.ts
```diff
@@ -60,7 +60,7 @@ export function recursiveFilesIn(directoryPath: string): Promise<string[]> {
 
 export function stat(filePath: string): Promise<fs.Stats> {
```
shockone: Could you rename this function to `lstat`?

test/utils/common_spec.ts
```diff
@@ -1,11 +1,27 @@
...
     });
+
+    describe("stat", () => {
+        it("returns stats even if the file is a borken symlink", async() => {
+            mockFs({
+                '/broken-symlink': mockFs.symlink({
+                    path: 'non-existing-file'
+                })
+            });
+            return stat(<FullPath>'/broken-symlink')
```
shockone: Please use `await` instead of `then`. You can use `resolveDirectory` to avoid type casting.

package.json
shockone: Unfortunately, typings aren't always up to date, but event the older ones are usually good enough.

viktomas:
@shockone: Thank you for the feedback.

shockone:
@viktomas thank you. May I ask you whether you find the codebase hard to understand, and what can be improved?

PR #567, opened 2016-07-07
Added show command which can display images inline in the terminal

Vatyx:
Currently there is no padding around the images which means that they are pressed up against each other. Also it seems like images don't get shrank either.

### DIFF
--- a/src/plugins/Show.tsx
+++ b/src/plugins/Show.tsx
@@ -1,24 +1,15 @@
-import * as _ from "lodash";
 import * as React from "react";
-import {homeDirectory, pluralize, resolveDirectory, resolveFile, mapObject} from "../utils/Common";
-import {existsSync, statSync} from "fs";
 import {PluginManager} from "../PluginManager";
 import {Job} from "../shell/Job";
 
 PluginManager.registerOutputDecorator({
     decorate: (job: Job): React.ReactElement<any> => {
-        const rows = job.screenBuffer.toLines().map(row => {
-            let fullPath: string = "hello";
-
-            fullPath = job.environment.cdpath.map(path => resolveDirectory(path, row))[0].slice(0, -1);
-
-            return <img src={fullPath}/>;
-        });
+        const rows = job.screenBuffer.toLines().map(path => <img src={path}/>);
 
         return <pre className="output">{rows}</pre>;
     },
 
     isApplicable: (job: Job): boolean => {
-        return job.hasOutput() && (job.prompt.expanded.map(token => token.value)[0] === "show");
+        return job.hasOutput() && (job.prompt.commandName.value === "show");
     },
 });
--- a/src/plugins/Show.tsx
+++ b/src/plugins/Show.tsx
@@ -1,0 +1,24 @@
+import * as _ from "lodash";
+import * as React from "react";
+import {homeDirectory, pluralize, resolveDirectory, resolveFile, mapObject} from "../utils/Common";
+import {existsSync, statSync} from "fs";
+import {PluginManager} from "../PluginManager";
+import {Job} from "../shell/Job";
+
+PluginManager.registerOutputDecorator({
+    decorate: (job: Job): React.ReactElement<any> => {
+        const rows = job.screenBuffer.toLines().map(row => {
+            let fullPath: string = "hello";
+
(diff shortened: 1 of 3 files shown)


src/shell/Command.ts
```diff
@@ -78,6 +78,9 @@ const executors: Dictionary<(i: Job, a: string[]) => void> = {
             throw `Don't know what to do with ${args.length} arguments.`;
         }
     },
+    show: (job: Job, args: string[]): void => {
+        job.screenBuffer.writeMany(args.join("\r\n"));
```
shockone: `"\r\n"` should be `os.EOS`. https://nodejs.org/api/os.html#os_os_eol

src/plugins/Show.tsx
```diff
@@ -0,0 +1,24 @@
...
+            fullPath = job.environment.cdpath.map(path => resolveDirectory(path, row))[0].slice(0, -1);
+
+            return <img src={fullPath}/>;
+        });
+
+        return <pre className="output">{rows}</pre>;
+    },
+
+    isApplicable: (job: Job): boolean => {
+        return job.hasOutput() && (job.prompt.expanded.map(token => token.value)[0] === "show");
```
shockone: Should be `job.prompt.commandName.value`.

src/plugins/Show.tsx
```diff
@@ -0,0 +1,24 @@
...
+import {existsSync, statSync} from "fs";
+import {PluginManager} from "../PluginManager";
+import {Job} from "../shell/Job";
+
+PluginManager.registerOutputDecorator({
+    decorate: (job: Job): React.ReactElement<any> => {
+        const rows = job.screenBuffer.toLines().map(row => {
+            let fullPath: string = "hello";
+
+            fullPath = job.environment.cdpath.map(path => resolveDirectory(path, row))[0].slice(0, -1);
```
shockone: `CDPATH` shouldn't be used for anything but `cd`. https://www.gnu.org/software/bash/manual/bashref.html#Bourne-Shell-Variables

src/shell/Command.ts
shockone: It would be better to resolve paths here, meaning to print full paths.

src/plugins/Show.tsx
```diff
@@ -0,0 +1,24 @@
+import * as _ from "lodash";
+import * as React from "react";
+import {homeDirectory, pluralize, resolveDirectory, resolveFile, mapObject} from "../utils/Common";
+import {existsSync, statSync} from "fs";
+import {PluginManager} from "../PluginManager";
+import {Job} from "../shell/Job";
+
+PluginManager.registerOutputDecorator({
+    decorate: (job: Job): React.ReactElement<any> => {
+        const rows = job.screenBuffer.toLines().map(row => {
```
shockone: Had you resolved paths in the command itself this could be just `const rows = job.screenBuffer.toLines().map(path => )`

src/plugins/Show.tsx
```diff
@@ -0,0 +1,24 @@
+import * as _ from "lodash";
+import * as React from "react";
+import {homeDirectory, pluralize, resolveDirectory, resolveFile, mapObject} from "../utils/Common";
+import {existsSync, statSync} from "fs";
+import {PluginManager} from "../PluginManager";
+import {Job} from "../shell/Job";
```
shockone: Please remove unused imports; you can check if you're find using `npm run lint`.

src/shell/Command.ts
```diff
@@ -78,6 +78,13 @@ const executors: Dictionary<(i: Job, a: string[]) => void> = {
             throw `Don't know what to do with ${args.length} arguments.`;
         }
     },
+    show: (job: Job, args: string[]): void => {
+        let imgs: string[] = [];
+        args.forEach(argument => {
+            imgs.push(resolveDirectory(job.environment.pwd, argument).slice(0, -1));
```
shockone: There is a `resolveFile` function to avoid a slash at the end. Also, `const images = args.map(...)` would be better.

shockone:
Thank you, @Vatyx. A new version will be automatically built and will appear here in a moment. https://github.com/shockone/black-screen/releases

PR #580, opened 2016-07-08
Add split panes feature

denieler:
Add feature to have opportunity to make splitting for panes horizontally and vertically. 

I've used same approach that you've chosen with grid-template-areas. Supporting any count of splitting in any direction. Have fun! 👍  

screenshot


denieler:
Hi, @shockone !

Here is the feature that was described in [this issue](https://github.com/shockone/black-screen/issues/372).

src/views/TabComponent.tsx
```diff
@@ -47,14 +47,39 @@ export class TabComponent extends React.Component<TabProps, TabState> {
 
 export class Tab {
     public sessions: Session[] = [];
+    /*
+    session view map has containing all sessions positions in format:
+    [ 1 4 5], where each number means count of columns in corresponding row
+     */
+    public sessionsViewMap: number[] = [
+      1,
+    ];
```
shockone: I'm afraid it should be a tree, not a 2d array. For example, if you split a window first vertically and then horizontally in iTerm, the horizontal split will only affect the right pane, whereas with this implementation it's split the whole screen.

src/shell/Session.ts
```diff
@@ -22,7 +22,7 @@ export class Session extends EmitterWithUniqueID {
     history = History;
     historicalPresentDirectoriesStack = new OrderedSet<string>();
 
-    constructor(private application: ApplicationComponent, private _dimensions: Dimensions) {
+    constructor(private application: ApplicationComponent, private _dimensions: Dimensions, public position: Positions) {
```
shockone: It's not correct to store position on the "model" for two reasons:

1. Position can change during the lifetime of a session. Like if you create three rows, and then delete the second one, the third row's position should change.
2. It's really view-related data.

shockone:
It seems that removing a session isn't handled: the rest of the sessions don't occupy the space released by a deleted one.

src/views/TabComponent.tsx
denieler: Yeah, you are right..

src/shell/Session.ts
denieler: where do you think the best place to store that? in `SessionComponent`, correct?

denieler:
@shockone what is the shortcut for removing session? I thought it was not implemented yet

shockone:
It's ctrl+d.

src/shell/Session.ts
shockone: I was thinking of storing sessions not as an array, but as a tree. And then during a tree traversal we'd know it's position.

denieler: @shockone actually you need to separate `TabComponent` from the `TabHeaderComponent` that pushed me add position here.. because it was only one place to add it safety

shockone:
@denieler, I see you're working on it. Please let me know when I should take a look at it.

denieler:
@shockone ok, sure

denieler:
@shockone hi, man! Please look on functionality.. I'll refactor it later, but need your approve about how it works now

denieler:
this test is not failing in my system, by the way )

denieler:
screen shot 2016-07-11 at 18 15 31

denieler:
screen shot 2016-07-11 at 19 41 49

shockone:
Closing a pane still doesn't seem to work. If I press Ctrl+D, it becomes not activateable, but is still rendered.

denieler:
@shockone yeah, that's true, but other stuffs are ok? and what I could do to fix the test?

shockone:
This issue with tests is fixed in https://github.com/shockone/black-screen/commit/d71072c4807ee3a209f87ddb7f3e15387adb0bc3, so you can just merge master. 

One small issue I've found is that the vertical border breaks at the top.

denieler:
@shockone added feature to remove splitted panes and fixed the bug with border, check please

shockone:
@denieler, it works. One more thing, though. When I switch between panes using cmd+j and cmd+k it navigates them in order of creation, not in order how they're displayed. To reproduce:

1. Create 4 horizontal splits.
2. Activate the top one.
3. Create one more horizontal split.
4. Navigate.

Expected result: the second split should be activated after the first one.
Actual result: the second split is activated after the 4th one.

denieler:
@shockone I totally agree with you, but could we fix this in new pull request?

In this pull request I'm planning to make some refactoring to make code more beautiful

shockone:
Yes, we can.

denieler:
it shouldn't be so much difficult, because we just need to change switching tabs from sessions to viewMapLeafs

src/views/1_ApplicationComponent.tsx
```diff
@@ -47,15 +48,15 @@ export class ApplicationComponent extends React.Component<{}, {}> {
 
     handleKeyDown(event: KeyboardEvent) {
         if (event.metaKey && event.keyCode === KeyCode.Underscore) {
-            this.activeTab.addSession();
+            this.activeTab.addSessionToPosition("vertical");
```
shockone: Please make sure you're using an enum for that.

denieler:
@shockone look on code please

src/utils/ViewMapLeaf.ts
```diff
@@ -0,0 +1,98 @@
+class ViewMapLeaf<T> {
+  public containerType?: ContainerType;
+  public parent: ViewMapLeaf<T>;
+  public childs: ViewMapLeaf<T>[];
```
shockone: Should be children.

denieler: what?

src/utils/ViewMapLeaf.ts
```diff
@@ -0,0 +1,98 @@
...
+    return previousValue;
+  }
+}
+
+enum ContainerType {
+  Row = 1,
+  Column = 2
+}
+
+export { ViewMapLeaf, ContainerType };
```
shockone: For consistency sake, please add the export modifiers directory to the classes.

src/utils/ViewMapLeaf.ts
denieler: `childs` show plurality to developers

src/utils/ViewMapLeaf.ts
denieler: ok

src/views/1_ApplicationComponent.tsx
```diff
@@ -133,37 +134,54 @@ export class ApplicationComponent extends React.Component<{}, {}> {
...
-
-                if (isActive) {
-                    return [sessionComponent];
-                } else {
-                    return [
-                        sessionComponent,
-                        <div className="shutter" style={css.sessionShutter(index)}></div>,
-                    ];
-                }
+        let renderSessionContainer: any;
```
shockone: I think a much better approach would be to just render a plain list of sessions, as it was before and position them using CSS grid areas.

denieler: @shockone we will create bicycle if will do that. Browser is already supporting tree structures using html, why we need to create our own full rendering processor?

src/utils/ViewMapLeaf.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  private value?: T;
+
+  constructor () {
+    this.childs = [];
+  }
+
+  public isContainer (): boolean {
+    return this.getValue() === undefined;
+  }
```
shockone: It seems that you've tried to combine both a branch and a leaf in a single class. Splitting them into `Branch` and `Leaf` would bring a couple of benefits:

* The name wouldn't be misleading: calling a branch a leaf is at least confusing.
* You would avoid optional properties
* You could have used polymorphism instead of checking `isContainer`.

src/views/1_ApplicationComponent.tsx
```diff
@@ -133,37 +134,54 @@ export class ApplicationComponent extends React.Component<{}, {}> {
...
+        let renderSessionRecursively: any;
+        let renderSession: any;
```
shockone: Please try to avoid `any`s.

src/utils/ViewMapLeaf.ts
denieler: yes, agree

src/views/css/main.ts
```diff
@@ -253,39 +243,80 @@ export const statusBar = {
...
+    top: 0,
+    bottom: 0,
+    left: 0,
+    right: 0,
+};
+
+export const jobsContainer = {
+  height: "100%",
+  width: "100%",
+  overflowY: "scroll",
```
shockone: I'm not sure I understand the purpose of this.

src/views/1_ApplicationComponent.tsx
denieler: couldn't here because Typescript is not allowing in `func1` use `func2` in same time when you use `func2` in `func1`

src/views/css/main.ts
```diff
@@ -253,39 +243,80 @@ export const statusBar = {
...
+
+  return styles;
+};
+
+export const sessionShutter = {
+    backgroundColor: colors.white,
+    zIndex: 1,
+    opacity: 0.2,
+    pointerEvents: "none",
+    position: "absolute",
```
shockone: Absolute positioning is something I tried to avoid. That's why I rendered them at the same level as sessions: for shutters to be in the same grid.

src/views/css/main.ts
denieler: put `shutter` inside `Session`, because it is related to `Session` actually. So `shutter` will gets all the space of `Session` to cover. But `jobsContainer` leaves still scrollable

src/views/css/main.ts
denieler: but they are related to `Session` component as for me, so should be in `Session` but not as  independent parts of nothing

src/views/1_ApplicationComponent.tsx
shockone: So, all you need to do in order to render them is to map the tree into a grid of areas. Like
```
Horizontal
  Leaf
  Leaf
  Vertical
    Leaf
    Leaf
```
would become

```
'session1 session1'
'session2 session2'
'session3 session4'
```

src/utils/ViewMapLeaf.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  public setValue (value: T | undefined) {
```
shockone: It would be much better to set data during initialization.

src/views/1_ApplicationComponent.tsx
denieler: imagine if will be more nested structure.. no, I totally don't like that, it will be ugly

src/views/1_ApplicationComponent.tsx
shockone: Probably you can move those render functions to the `Leaf` and `Branch` classes I mentioned above.

src/views/1_ApplicationComponent.tsx
shockone: This mapping will still be necessary in the future to know what session is located, say, above current. But we can delay it, if you don't feel like doing that.

src/views/2_SessionComponent.tsx
```diff
@@ -35,14 +35,23 @@ export class SessionComponent extends React.Component<Props, {}> {
                           hasLocusOfAttention={this.props.isActive && index === this.props.session.jobs.length - 1}/>
         );
 
+        let shutter: any;
+        if (!this.props.isActive) {
+          shutter = <div className="shutter" style={css.sessionShutter}></div>;
+        }
```
shockone: Let's simplify this flow and always render the shutter, but change it's opacity based on `isActive`.

src/views/1_ApplicationComponent.tsx
denieler: for which features? now we have tree that manage all rendering and could know about everything from this info

shockone: In iTerm2, I can press a shortcut to navigate to a visually neighbouring pane.

src/views/css/main.ts
shockone: Got it.

src/views/css/main.ts
shockone: Agreed.

src/utils/ViewMapLeaf.ts
shockone: I don't insist on changing it, but I've never seen such word before.

http://docs.oracle.com/javase/7/docs/api/javax/swing/tree/TreeNode.html#children()
https://api.jquery.com/children/

src/views/1_ApplicationComponent.tsx
denieler: @shockone I thought a little about this feature.. and actually it is not seems for me that to make this feature based on `display: grid` much easier that based on the tree.. complexity is the same actually.

shockone: I don't know all the peculiarities of this one, but I believe you. :)

src/utils/ViewMapLeaf.ts
denieler: @shockone hey! I've tried today move `ViewMapLeaf` to `ViewMapLeaf` and `ViewMapBranch`and to be honest it looks super weird in the code. Typescript is not supporting now command `instanceof` for generic classes, so it is first weird place, another place is when we are finding recursively we need to have methods for each type and return only `ViewMapLeaf`.. so complexity is much bigger, readability much better.. so I decided to leave it now as it is.

src/views/2_SessionComponent.tsx
denieler: @shockone this is done, actually, very nice idea, thanks!

shockone:
Do you plan to make more changes, or can I take a look again?

denieler:
Yes, actually I'm planning to leave it for now, because don't have opportunity to spend more time to this feature now

denieler:
In two weeks maybe I can add moving by split panes, if you'll not finish it until that time

shockone:
Got it, thank you for your hard work. It's the biggest and the most useful pull request ever for Black Screen!

denieler:
@shockone thank you ) sorry for strange code sometimes ) If I'll try to fix `ViewMapLeaf` properly - I'll push the code for sure

PR #591, opened 2016-07-10
Fixes deprecated app-bundle-id

### LINKED ISSUE #590: Remove deprecated app-bundle-id
On package.json, under `build`, we're using `app-bundle-id`, which is deprecated by Electron Builder. We should use `appId` instead.

lbalves:
Uses appId instead
Fixes #590

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "app": "."
   },
   "build": {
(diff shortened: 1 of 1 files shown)


shockone:
Cool, thanks, @lbalves.

PR #594, opened 2016-07-10
Autocomplete for ls

drew-gross:
I'm experimenting with generating autocomplete from man pages. This is a potential output. The content is from the man pages, IDK if using it like this jives with your license or not.

### DIFF
--- a/src/plugins/autocompletion_providers/Ls.ts
+++ b/src/plugins/autocompletion_providers/Ls.ts
@@ -5,202 +5,167 @@ import {mapObject} from "../../utils/Common";
 const optionsProvider = mapObject(
     {
         "@": {
-            short: "",
             long: "Display extended attribute keys and sizes in long (-l) output.",
         },
 
         "1": {
-            short: "",
             long: "The numeric digit ``one''.)  Force output to be one entry per line.  This is the default when output is not to a terminal.",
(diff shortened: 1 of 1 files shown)


src/plugins/autocompletion_providers/Ls.ts
```diff
@@ -0,0 +1,206 @@
...
+            long: "Force raw printing of non-printable characters.  This is the default when output is not to a terminal.",
+        },
+
+        x: {
+            short: "",
+            long: "The same as -C, except that the multi-column output is produced with entries sorted across, rather than down, the columns.",
+        },
+
+    },
+    (option, descriptions) => mapSuggestions(shortFlag(option), suggestion => suggestion.withSynopsis(descriptions.short).withDescription(descriptions.long))
```
shockone: May I ask you to remove the `short` keys from all the objects together with the `.withSynopsis(descriptions.short)` part? They don't really add any value.

By the way, how hard was it to add those?

drew-gross: Done, as well as shortening lines to fix lint.

It was pretty easy to add these.

RobertoUa:
I think that ideally parsing of man pages should be done users computer.

PR #602, opened 2016-07-12
Prevent autocomplete popup when navigate by history

ekaragodin:


shockone:
Awesome! By the way, you can use Ctrl+N and Ctrl+P to navigate.

PR #607, opened 2016-07-13
Improve git autocomplete

ekaragodin:


src/plugins/autocompletion_providers/Git.ts
```diff
@@ -50,9 +51,65 @@ const configOptions = combine([
...
+    "force",
+    "keep",
+    "depth=",
+    "tags",
+    "no-tags",
+    "all",
+    "prune",
+    "dry-run",
+    "recurse-submodules=",
+].map(option => longFlag(option)));
```
shockone: This is the same as `.map(longFlag)`.

src/plugins/autocompletion_providers/Git.ts
```diff
@@ -50,9 +51,65 @@ const configOptions = combine([
...
+    }
+
+    return [];
+};
+
 const configVariables = unique(async(context: AutocompletionContext): Promise<Suggestion[]> => {
     const variables = await Git.configVariables(context.environment.pwd);
-    return variables.map((variable) => new Suggestion().withValue(variable).withStyle(styles.option));
+
+    return variables.map((variable) => {
```
shockone: There is shorter syntax for it: `map(variable => new Suggestion().withThings())`

src/plugins/autocompletion_providers/Git.ts
```diff
@@ -84,181 +141,310 @@ const notStagedFiles = unique(async(context: AutocompletionContext): Promise<Sug
...
-    commit: commitOptions,
-    status: statusOptions,
-    merge: combine([branchesExceptCurrent, branchAlias]),
-    push: pushOptions,
-    config: combine([configVariables, configOptions]),
-};
+interface GitCommandData {
+    name: string;
+    description: string;
+    provider?: AutocompletionProvider;
```
shockone: It would be better to create an `empty` provider and use it everywhere than having an optional property.

src/plugins/autocompletion_providers/Git.ts
```diff
@@ -84,181 +141,310 @@ const notStagedFiles = unique(async(context: AutocompletionContext): Promise<Sug
...
+                if (!data) {
+                    subCommandsData.push({
+                        name,
+                        description: "",
+                        aliases: [],
+                    });
+                }
+            });
+    }
+}
```
shockone: Oh, man. In place modification is so not cool. I'd much rather have `combine([staticSubcommands, dynamicSubcommands])` with some performance penalty that this.

src/plugins/autocompletion_providers/Git.ts
```diff
@@ -84,181 +141,310 @@ const notStagedFiles = unique(async(context: AutocompletionContext): Promise<Sug
...
+                data.aliases.push(name);
+            } else {
+                subCommandsData.push({
+                    name: name,
+                    description: variable.value,
+                    aliases: [],
+                });
+            }
+        });
+}
```
shockone: Same here. Just add it to the aforementioned combine. Data mutation is one of the most widespread cause of bugs.

src/plugins/autocompletion_providers/Git.ts
```diff
@@ -84,181 +141,310 @@ const notStagedFiles = unique(async(context: AutocompletionContext): Promise<Sug
...
+    aliases: string[];
+}
 
-const subCommands = [
+const subCommandsData: GitCommandData[] = [
     {
         name: "add",
         description: "Add file contents to the index.",
+        provider: combine([notStagedFiles, addOptions]),
+        aliases: [],
```
shockone: Why do you need to associate aliases with particular subcommands? It seems like it doesn't matter much.

src/plugins/autocompletion_providers/Git.ts
```diff
@@ -84,181 +141,310 @@ const notStagedFiles = unique(async(context: AutocompletionContext): Promise<Sug
...
+                    .withValue(name)
+                    .withDescription(data.description)
+                    .withStyle(styles.command)
+                    .withSpace();
+
+                result.push(aliasSuggestion);
+            });
+
+            return result;
+        });
```
shockone: This one should be `combine([staticSubcommands, dynamicSubcommands, aliases])`.

src/plugins/autocompletion_providers/Git.ts
```diff
@@ -84,181 +141,310 @@ const notStagedFiles = unique(async(context: AutocompletionContext): Promise<Sug
...
         if (firstArgument) {
-            const subCommandProvider = subCommandProviders[firstArgument.value];
-            if (subCommandProvider) {
-                return subCommandProvider(context);
+            const data = find(subCommandsData, (data) => {
+                if (data.name === firstArgument.value) {
+                    return true;
+                }
+
+                return data.aliases.indexOf(firstArgument.value) !== -1;
```
shockone: Ah, now I see why you want to associate aliases with subcommands.

shockone:
I wrote some feedback, but I'll merge it even if you don't change anything, because it's awesome!

ekaragodin:
I want make some performance optimizations before merge

shockone:
@ekaragodin, so, do you plan to work on it some more, or is it finished?

ekaragodin:
Currently I finished

shockone:
It's so good, @ekaragodin! Merging right now.

shockone:
For your information, I've changed a few things:

* Moved `aliases` before `commands` in `allCommands` because they seem more important.
* Created a helper function `contextIndependent`, which does memoization and forces the provider not to depend on context.
* Moved `emptyProvider` to `Common.ts`.

PR #610, opened 2016-07-13
Load both ~/.bash_history and ~/.black-screen/history

drew-gross:
Due to some bugs in black-screen I switch between iTerm2 and black-screen. This change reads the command history from both history files, putting the one that was written to most recently first.

### DIFF
--- a/src/shell/Session.ts
+++ b/src/shell/Session.ts
@@ -113,7 +113,6 @@ export class Session extends EmitterWithUniqueID {
     };
 
     private deserialize(): void {
-        debugger;
         this.directory = this.readSerialized(presentWorkingDirectoryFilePath, homeDirectory);
         let bashHistoryModifiedTime = new Date(0);
         let bashHistory: string[] = [];
--- a/src/shell/Session.ts
+++ b/src/shell/Session.ts
@@ -118,7 +118,7 @@ export class Session extends EmitterWithUniqueID {
         let bashHistory: string[] = [];
         try {
             bashHistoryModifiedTime = statSync(bashHistoryFilePath).mtime;
-            bashHistory = readFileSync(bashHistoryFilePath).toString().trim().split('\n').reverse();
+            bashHistory = readFileSync(bashHistoryFilePath).toString().trim().split("\n").reverse();
         } catch (error) {
             // File was missing or something. Ignore.
         }
@@ -132,7 +132,7 @@ export class Session extends EmitterWithUniqueID {
             // File was missing or something. Ignore.
         }
 
-        let mergedHistory: string[] = []
+        let mergedHistory: string[] = [];
         if (blackScreenHistoryModifiedTime > bashHistoryModifiedTime) {
             mergedHistory = [...bashHistory, ...blackScreenHistory];
         } else {
--- a/src/shell/Session.ts
+++ b/src/shell/Session.ts
@@ -1,4 +1,4 @@
-import {readFileSync, statSync} from "fs";
+import {readFileSync} from "fs";
 import {Job} from "./Job";
 import {History} from "./History";
 import {EmitterWithUniqueID} from "../EmitterWithUniqueID";
@@ -9,7 +9,6 @@ import {Environment, processEnvironment} from "./Environment";
 import {
     homeDirectory, normalizeDirectory, writeFileCreatingParents,
     presentWorkingDirectoryFilePath, historyFilePath,
-    bashHistoryFilePath,
 } from "../utils/Common";
 import {remote} from "electron";
 import {OrderedSet} from "../utils/OrderedSet";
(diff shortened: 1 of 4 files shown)


ekaragodin:
It would be great to support zsh_history

shockone:
Hey, @drew-gross. It might be a good short-term fix, but I'd rather not do it for two reasons:

* We plan to extend the history format to include start time, end time, session id, PWD and maybe something else. It'll make them incompatible.
* We should really make it stable instead of fixing the symptoms.

shockone:
Although, we might leave it for now.

src/utils/Common.ts
```diff
@@ -350,4 +350,6 @@ export function assign<A extends B, B extends Object>(source: A, assignments: B)
 const baseConfigDirectory = Path.join(homeDirectory, ".black-screen");
 export const presentWorkingDirectoryFilePath = Path.join(baseConfigDirectory, "presentWorkingDirectory");
 export const historyFilePath = Path.join(baseConfigDirectory, "history");
+// We can't use HISTFILE to get the path, so just check the default location
+export const bashHistoryFilePath = Path.join(homeDirectory, ".bash_history");
```
shockone: There is a better place for that. https://github.com/shockone/black-screen/blob/master/src/utils/Shell.ts. You can also move deserialization logic there, if it differs across shells.

src/shell/Session.ts
```diff
@@ -112,7 +113,32 @@ export class Session extends EmitterWithUniqueID {
 
     private deserialize(): void {
         this.directory = this.readSerialized(presentWorkingDirectoryFilePath, homeDirectory);
-        History.deserialize(this.readSerialized(historyFilePath, []));
+        let bashHistoryModifiedTime = new Date(0);
```
shockone: I'd say this code belongs to `History.deserialize`.

src/shell/Session.ts
```diff
@@ -112,7 +113,32 @@ export class Session extends EmitterWithUniqueID {
...
+            blackScreenHistory = this.readSerialized(historyFilePath, []);
+        } catch (error) {
+            // File was missing or something. Ignore.
+        }
+
+        let mergedHistory: string[] = [];
+        if (blackScreenHistoryModifiedTime > bashHistoryModifiedTime) {
+            mergedHistory = [...bashHistory, ...blackScreenHistory];
+        } else {
+            mergedHistory = [...blackScreenHistory, ...bashHistory];
```
shockone: Would it be possible to interleave the records based on time? It's not trivial to do, so I'm OK with a "no".

src/shell/Session.ts
```diff
@@ -112,7 +113,32 @@ export class Session extends EmitterWithUniqueID {
 
     private deserialize(): void {
         this.directory = this.readSerialized(presentWorkingDirectoryFilePath, homeDirectory);
-        History.deserialize(this.readSerialized(historyFilePath, []));
+        let bashHistoryModifiedTime = new Date(0);
+        let bashHistory: string[] = [];
+        try {
+            bashHistoryModifiedTime = statSync(bashHistoryFilePath).mtime;
+            bashHistory = readFileSync(bashHistoryFilePath).toString().trim().split("\n").reverse();
```
shockone: `"\n"` should be `os.EOL`.

src/shell/Session.ts
drew-gross: As far as I can tell, the time is not stored in `.bash_history`.

drew-gross:
I hope we can also make this stable, but even once it is, I think loading their original shell history will still be nice, especially for people switching to black-screen for the first time.

src/shell/Session.ts
shockone: Actually, there is. Or rather you can configure the shell to have it. So, if not to support it, we need to at least verify the format we read is expected. The keyword to find more is `HISTTIMEFORMAT`.

drew-gross:
OK, I made the suggested changes.

src/utils/Shell.ts
```diff
@@ -78,4 +113,8 @@ const {shell} = new class {
     };
 };
 
+export function loadAllHistories(): { lastModified: Date, commands: string[] }[] {
```
shockone: You can safely assume that a person only wants to read history from his login shell, which is `loginShell`. Also, it'll be consistent with alias loading behaviour. So, please remove this function and use `loginShell. loadHistory()` instead.

src/utils/Shell.ts
```diff
@@ -28,6 +34,27 @@ class Bash extends Shell {
     get preCommandModifiers(): string[] {
         return [];
     }
+
+    loadHistory(): { lastModified: Date, commands: string[] } {
+        if (process.env.HISTTIMEFORMAT) {
```
shockone: This check is pretty much useless: Black Screen will only pick up shell's environment if it's started from a shell session. But usually it's started as a standalone application.

drew-gross:
Done.

PR #613, opened 2016-07-15
Clean up code for finding user's shell

drew-gross:
I noticed the shell() function was memoizing, which seems unnecessary since it's not an expensive operation and is only ever called once anyway. I also removed `baseName`, which seems to be a re-implementation of built in function `basename`.


shockone:
👍

PR #616, opened 2016-07-17
Removed ligatures from prompt

### LINKED ISSUE #615: The command line has unwanted ligatures in it
Here's how it looks in action:

screen shot 2016-07-17 at 13 04 09

I didn't gather debug logs because I don't think they are necessary.

bananu7:
This fixes #615.

Another option would be to relax the type definition for that field to allow only some ligatures to be disabled.

### DIFF
--- a/src/views/css/definitions.ts
+++ b/src/views/css/definitions.ts
@@ -51,7 +51,6 @@ export interface CSSObject {
     textDecoration?: "underline";
     fontWeight?: "bold";
     fontSize?: number;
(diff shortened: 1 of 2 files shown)


DeadMG:
The React.d.ts file should really be committed to the repository.

shockone:
Please remove this line as well. https://github.com/shockone/black-screen/blob/d1d91d18a25aa3e7b4387dad775e976ae4b9fcb1/src/views/css/definitions.ts#L54-L54

bananu7:
Done.

PR #617, opened 2016-07-17
Autocomplete vagrant

ekaragodin:


src/plugins/autocompletion_providers/Vagrant.ts
```diff
@@ -0,0 +1,40 @@
...
+import {styles, Suggestion, contextIndependent} from "./Common";
+
+const commands = contextIndependent(async() => {
+    let lines: string[] = [];
+
+    try {
+        lines = await linedOutputOf("vagrant", ["list-commands"], process.env.HOME);
+    } catch (e) {
+        // skip if command is not exist
+    }
```
shockone: Using exceptions as flow control is an antipattern. http://programmers.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why

src/plugins/autocompletion_providers/Vagrant.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    return lines
+        .map(line => {
+            const matches = line.match(/([\-a-zA-Z0-9]+)  /);
+
+            if (matches) {
+                const name = matches[1];
+                const description = line.replace(matches[1], "").trim();
+
+                return new Suggestion()
```
shockone: Please use the new API, where you'd pass everything into the constructor. You can find examples in any other provider.

src/plugins/autocompletion_providers/Vagrant.ts
```diff
@@ -0,0 +1,40 @@
...
+                    .withValue(name)
+                    .withDescription(description)
+                    .withStyle(styles.command)
+                    .withSpace();
+            }
+
+            return undefined;
```
shockone: This return isn't necessary here. If you don't return a value, `undefined` will be returned anyway.

src/plugins/autocompletion_providers/Vagrant.ts
```diff
@@ -0,0 +1,40 @@
...
+import {linedOutputOf} from "../../PTY";
+import {styles, Suggestion, contextIndependent} from "./Common";
+
+const commands = contextIndependent(async() => {
+    let lines: string[] = [];
+
+    try {
+        lines = await linedOutputOf("vagrant", ["list-commands"], process.env.HOME);
+    } catch (e) {
+        // skip if command is not exist
```
shockone: There is a function `executablesInPaths` to check if you have this executable. It requires `EnvironmentPath`, though, which makes it not `contextIndependent` anymore. But it's OK.

ekaragodin:
done

shockone:
👍

PR #621, opened 2016-07-19
Add autocomplete for mkdir (by parsing man page)

drew-gross:
Eventually I will split out the man page parsing stuff into a more reusable place, but for now it only works on the `mkdir` man page. I plan to handle more and more man pages over time and hopefully eventually handle all man pages by default.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
     "fuzzaldrin": "2.1.0",
     "immutable": "3.8.1",
     "lodash": "4.13.1",
-    "manolo": "0.1.0",
     "mode-to-permissions": "0.0.2",
     "node-ansiparser": "2.2.0",
     "pty.js": "shockone/pty.js",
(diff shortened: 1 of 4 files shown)


shockone:
Wow, that's a big undertaking. You might be interested to see how it's done in the Fish shell https://github.com/fish-shell/fish-shell/blob/master/share/tools/create_manpage_completions.py. Also, the format of man files is `troff` or its dialects, and there are existing JS parsers for it. I'm not sure how good they are though.

One thing I'd ask you is to keep all the man parser related things in one file, if it's possible.

shockone:
Currently the `-m` completion misbehaves a bit. It completes the word `mode` itself too, even though it's not what a person would want to type. There is a special property of `Suggestion` called `displayValue`, which you can set to `-m mode`, and then set `value` to just `-m` in order to have a better behaviour.

drew-gross:
Thanks for the tips. I will fix "-m" and look at that `fish` code.I tried a few of the existing `troff` parsers, but they all didn't work at all (as in they threw away most of the input). The man pages seem to include some non-standard troff commands or something.

PR #629, opened 2016-07-21
Autocomplete rm by parsing man page

drew-gross:


src/utils/ManPages.ts
```diff
@@ -0,0 +1,98 @@
...
+            } else {
+                return memo.concat(next, " ");
+            }
+        },
+        ""
+    );
+
+// Note: this is still pretty experimental. If you want to do man page parsing
+// for a new command, expect to have to make some changes here.
+export let manPageToOptions = (command: string): Promise<AutocompletionProvider> => {
```
shockone: Please use `const` instead of `let` by default. Only use `let` if you have to.

shockone: You can as well return `AutocompletionProvider` from this function.

src/utils/ManPages.ts
```diff
@@ -0,0 +1,98 @@
...
+    // use execFile to prevent a command like "; echo test" from running the "echo test"
+    return new Promise((resolve, reject) => {
```
shockone: Please use `async/await` instead of promises. You'd need to create a promisified wrapper for `execFile` in `utils/Common.ts`. There are many such wrappers to copy from.

drew-gross:
Done.

shockone:
Could you please update it? I can't merge.

drew-gross:
Done. I went ahead and merged it since it seems you've OKed it. Hope thats OK.

shockone:
Sure, Drew. :)

PR #636, opened 2016-07-22
Use ~/.profile file (fixes #618)

### LINKED ISSUE #618: Bash alises and functions from ~/.profile don't work.
image

In my `~/.profile` file I have (among other things)

```bash
alias l="ls -lGhpa"

function gc() {
    git commit $@ && git status
}
export -f gc
```

Bash version is 4.3.46 (`GNU bash, version 4.3.46(1)-release (x86_64-apple-darwin15.5.0)`)

`black-screen` version is master branch.

I'm happy to help debug and fix this, I've been looking at it a little already, but I'm not super familiar with the codebase or pty stuff, so if you have some pointer for where to start or ideas about the cause I would be very glad.

drew-gross:
Bash (at least on mac) loads from ~/.profile as well

### DIFF
--- a/src/utils/Shell.ts
+++ b/src/utils/Shell.ts
@@ -38,7 +38,7 @@ class Bash extends Shell {
     }
 
     get configFiles() {
-        return ["~/.bashrc", "~/.bash_profile"];
+        return ["~/.bashrc", "~/.bash_profile", "~/.profile"];
     }
 
     get noConfigSwitches() {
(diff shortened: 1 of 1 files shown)


shockone:
Could you please find a list of all files both shells load *for interactive sessions* in their documention and add all of them here to close it once and forever?

drew-gross:
Done. Used the list from each shell's man page, from the one I have on my system. I haven't tried the zsh one since I don't use zsh.

shockone:
You're awesome, man.

PR #640, opened 2016-07-23
Strip bash funcs

drew-gross:
The removes all `BASH_FUNC_foo%%` env vars at load time, as those cause issues when actually running bash commands. I couldn't find a better place to put this, if you have a suggestion I'd be happy to move it there.

### DIFF
--- a/src/shell/Environment.ts
+++ b/src/shell/Environment.ts
@@ -8,6 +8,17 @@ import {AbstractOrderedSet} from "../utils/OrderedSet";
 const ignoredEnvironmentVariables = [
     "NODE_ENV",
 ];
+
+const isIgnoredEnvironmentVariable = (varName: string) => {
+    if (ignoredEnvironmentVariables.includes(varName)) {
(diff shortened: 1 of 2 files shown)


src/shell/Environment.ts
```diff
@@ -15,7 +15,9 @@ export async function loadEnvironment(): Promise<void> {
     lines.forEach(line => {
         let [key, value] = line.trim().split("=");
 
-        if (!ignoredEnvironmentVariables.includes(key)) {
+        // Strip bash functions from environment, as they cause issues.
+        // TODO: put the code for stripping bash functions in a better place.
+        if (!ignoredEnvironmentVariables.includes(key) && !/^BASH_FUNC\w+%%$/.test(key)) {
```
shockone: Can we create a function `isIgnoredEnvironmentVariable`, and use it in this `if` instead of those two predicates?

drew-gross:
Done :)

PR #641, opened 2016-07-23
Fix pasting when prompt already contains text

### LINKED ISSUE #635: Copy paste removes previously entered text
When first entering text

image

and then copy paste, the previously entered text will be removed

image

This happens with version 0.2.8 (0.2.8.364) on OSX 10.10.5.

drew-gross:
Fix by allowing the event to continue propagating when job is not in progress.

Fixes #635

### DIFF
--- a/src/views/4_PromptComponent.tsx
+++ b/src/views/4_PromptComponent.tsx
@@ -57,12 +57,9 @@ export class PromptComponent extends React.Component<Props, State> {
 
         if (this.props.status === e.Status.InProgress) {
             this.props.job.write(text);
-        } else {
-            this.setText(text);
+            event.stopPropagation();
+            event.preventDefault();
         }
-
-        event.stopPropagation();
-        event.preventDefault();
     };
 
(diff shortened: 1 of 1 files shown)


shockone:
If I blur the prompt (click somewhere on the output), I should be able to paste into the prompt as well, but with this change I won't be able to. Like currently you can type not thinking which element is focused, and the typed text will be sent to prompt anyway.

drew-gross:
It still seems to work that way for me, if I click anywhere but the search bar I can still type and have it go to the prompt

shockone:
Exactly, but can you *paste* and have it go to the prompt?

drew-gross:
Ah I see. I'll see if I can fix that too.

drew-gross:
Pasting when the prompt is blurred doesn't seem to work for me on `master` either.

shockone:
Huh, my bad. :) Then let's merge it and I'll fix it separately.

PR #645, opened 2016-07-24
Autocomplete for cat, ln, ls, pwd, shutdown

drew-gross:


RobertoUa:
Wow! We now have man-pages parsing. Amazing.
Although I think it shouldn't work like this. BS should parse them automatically (user writes a command, BS checks if it has a man page, if not - don't check again for another like week). And probably a shell command to rebuild completitions from history or just remove saved data.

shockone:
@RobertoUa, yeah, it's a good optimization, and it's what Fish does too. But it's still an optimization, and we can worry about it later.

src/utils/ManPages.ts
```diff
@@ -18,6 +17,14 @@ const combineManPageLines = (lines: string[]) => lines
...
 // Note: this is still pretty experimental. If you want to do man page parsing
 // for a new command, expect to have to make some changes here.
+
+// TODO: Handle option descriptions that have empty lines. Unblocks:
+// -p and -R in cp
+// df
+// locate
+// TODO: Handle nested options. Unblocks:
+// dd
+
```
shockone: I think we should start writing tests for this component.

RobertoUa:
You write auto parsing once instead of adding manual parsing only for some commands.

shockone:
Ah, I see now. We can pretty easily add a man autocompletion provider to all executables with one line of code, but also maybe a hand-written provider should supersede it, because it's supposedly of better quality. What do you think?

RobertoUa:
Sure, BS should support different levels of autocompletitions with defined priority.

drew-gross:
@RobertoUa currently the man page parser will fail on most man pages as it can only handle basic stuff. That's why I've only enabled it for commands where I've checked that it works. I agree that hand-written providers should supersede man pages, possibly even on an option-by-option level. The man pages parse is still really in early stages, I'm sure we'll make it work better as we go.

@shockone I'll start adding tests in the next PR.

PR #647, opened 2016-07-25
Add commands from history to autocomplete

drew-gross:
Show commands run in the past as autocomplete suggestions if their prefix matches what you've already typed.

image

image

image

### DIFF
--- a/src/Autocompletion.ts
+++ b/src/Autocompletion.ts
@@ -1,10 +1,19 @@
 import {Job} from "./shell/Job";
 import {leafNodeAt} from "./shell/Parser";
 import * as _ from "lodash";
+import {History} from "./shell/History";
+import {Suggestion, styles} from "./plugins/autocompletion_providers/Common";
 
 export const suggestionsLimit = 9;
 
 export const getSuggestions = async(job: Job, caretPosition: number) => {
+    const prefixMatchesInHistory = History.all.filter(line => line.startsWith(job.prompt.value));
+    const suggestionsFromHistory = prefixMatchesInHistory.map(match => new Suggestion({
+        value: match,
+        shouldEscapeSpaces: false,
+        style: styles.command,
+    }));
+
     const node = leafNodeAt(caretPosition, job.prompt.ast);
     const suggestions = await node.suggestions({
         environment: job.environment,
@@ -12,7 +21,7 @@ export const getSuggestions = async(job: Job, caretPosition: number) => {
         aliases: job.session.aliases,
     });
 
-    const applicableSuggestions = _.uniqBy(suggestions, suggestion => suggestion.value).filter(suggestion =>
+    const applicableSuggestions = _.uniqBy([...suggestionsFromHistory, ...suggestions], suggestion => suggestion.value).filter(suggestion =>
         suggestion.value.toLowerCase().startsWith(node.value.toLowerCase())
     );
 
--- a/src/Autocompletion.ts
+++ b/src/Autocompletion.ts
@@ -11,8 +11,11 @@ export const getSuggestions = async(job: Job, caretPosition: number) => {
     const suggestionsFromHistory = prefixMatchesInHistory.map(match => new Suggestion({
         value: match,
         shouldEscapeSpaces: false,
-        style: styles.command,
+        style: styles.history,
     }));
+
+    const firstThreeFromHistory = suggestionsFromHistory.slice(0, 3);
+    const remainderFromHistory = suggestionsFromHistory.slice(3);
 
     const node = leafNodeAt(caretPosition, job.prompt.ast);
     const suggestions = await node.suggestions({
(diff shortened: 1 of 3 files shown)


shockone:
There was history completion in the older versions, but I removed it because it has proven to be too noisy. I planned to have it only in reverse history search, but if to add it here, we need a stricter criteria than `startsWith` to reduce noise.

drew-gross:
Would you merge this if I made it optional via a setting somewhere? I would use this as is.

shockone:
Hiding behind an option isn't an option because "as little configuration as possible" is one of my guiding principles.

In my opinion it should really be in reverse history search.

drew-gross:
Do you have an example of what you meant by too noisy? I'd really like to get this in, or at least some version of this. I agree in general with having as little configuration as possible but if there is a genuine tradeoff and sensible defaults (with useful features defaulting to enabled) then I think configuration options are fine. Sublime text does configuration really well, I think.

Personally I dislike reverse-history-search because (in bash at least) it includes infix matches and not just prefix matches, which is almost never what I want.

shockone:
By noisy I mean that there are so many permutations and combinations of a command, that had you executed `docker` from your example a couple more times with different arguments, history'd occupy the whole list of suggestions.

I'm sorry, but I don't think adding an options is a good solution. We need to decide what's better for a person (by the way, target audience is professionals who know their tools and want to be more productive).

The main insight here is that every person has a set of commands she repeat more often than others. But in the current implementation we don't distinguish highly-repeated commands from others: everything that is in history is equal and is automatically worth suggesting. Here's a command:

```shell
cat without_mappings.csv | awk 'BEGIN { FS = "," } { print $4 }'
```

 I've executed it once a month ago, and now every time I type `cat` I'll see this suggestion. I find that having negative effect on autocompletion quality.

On the other hand, I have commands that I execute at least a few times a day, like `git add -p`. These are not the same and shouldn't be treated equally.

I had two ideas in mind, which aren't mutually exclusive. Both of them require us to store more historical data than we do now.

* Suggest to create aliases for repeated commands. It could be a popover saying `You seem to use "git add -p" quite often. Wanna create an alias "gap" for it? <Yes> <No>`. I'm not sure how it will be accepted, because I had this plugin once, which would suggest me shortcuts when I executed an action using mouse, and it was quite annoying.

* Create a machine learning component, which would guess the command you want to execute. We could have the following [features](https://en.wikipedia.org/wiki/Feature_(machine_learning)) (basically things that might influence the result): text, pwd of the executed command, previous command, time from previous command, day (cause I tend to type different things on weekdays than on weekends) and let it figure out what I plan to do next based on historical data. Now it won't suggest me to execute `git add` in a directory where I never executed it before. Maybe it's not even a git directory after all. And if I always do `git checkout master` and then `git pull`, I'd get a suggestion for exactly what I want, not one of myriad of similar commands.

drew-gross:
What would you think about taking the most recent 1-3 commands that match the prefix, and adding them to front of the autocompletion suggestions, and adding the rest to the back of the autocompletion suggestions?

shockone:
As a temporary measure we can do it like this: if the history contains up to three commands with such prefix, we display them all. If more - we don't include the history suggestions at all because we aren't sure what a person wants. What do you say?

src/Autocompletion.ts
```diff
@@ -1,17 +1,26 @@
...
+import {Suggestion, styles} from "./plugins/autocompletion_providers/Common";
 
 export const suggestionsLimit = 9;
 
 export const getSuggestions = async(job: Job, caretPosition: number) => {
+    const prefixMatchesInHistory = History.all.filter(line => line.startsWith(job.prompt.value));
+    const suggestionsFromHistory = prefixMatchesInHistory.map(match => new Suggestion({
+        value: match,
+        shouldEscapeSpaces: false,
+        style: styles.command,
```
shockone: Please create a separate style for history. It used to have a blue icon with a clock: http://fontawesome.io/icon/history/

drew-gross:
That seems like it would end up being not very useful for commands that have sub-commands like `git`, `docker`, `brew`, `npm`, etc, and also it could cause people a lot of confusion to people if e.g. they have 3 commands they use very frequently like, say, `ssh <hostname1>`, `ssh <hostname2>`, `ssh <hostname3>`, and they frequently switch between the 3 hostnames, and then later they start using a 4th hostname and all of a sudden none of the 4 hostnames appear in autocomplete anymore.

drew-gross:
One thing we could consider is mimicking Xcode's autocomplete, which I think is pretty much the best implementation of auto complete. If you haven't used it, the way it works is that if you have typed `docke` and your suggestion list contains `docker` and `docker rmi -f $(docker images -q)`, then hitting `tab` would only fill in the prompt up until you have another "decision" to make, so in this case it would will in "r" and the space. But, if you select something with arrow keys, and then press `tab`, it would fill in the whole thing you selected.

drew-gross:
I added the history styling and added the change to make only the first 3 commands from history be part of the autocomplete. I'd like to keep the UI predictable though, and I don't really think hiding the history autocomplete after some uses a 4th matching command will make the UI predictable.

image

shockone:
It's a good point about UI predictability. I'll merge it but it's still considered experimental and is a subject to further change or improvements.

shockone:
Thank you for your work, @drew-gross. By the way, I'm going to be on vacation for about two weeks starting tomorrow, so I won't be available.

drew-gross:
Thanks for merging, I also hope to keep improving the autocomplete. Have fun on your vacation!

shockone:
Have you noticed the suggestions disappear if there is more than one token? It's because of the `startWith` check which is performed on the node under caret, not on the whole input. Maybe you could do something about it?

drew-gross:
Yeah I had noticed some weird behaviour, I guess that's the cause. I'm also on vacation, but not for two weeks, so I'll try to get to this on Tuesday or Wednesday.

PR #650, opened 2016-07-25
Move paste handling to handleUserEvent.

shockone:
Fixes https://github.com/shockone/black-screen/issues/643

@drew-gross please review as you modified this code before.


drew-gross:
Sweet, looks like this should work. Not sure why the tests are failing, but once those are fixed this should be fine to merge.

PR #676, opened 2016-08-07
Fix shutter for large sessions

drew-gross:
Previously, when you scroll a session, the shutter scrolled up as well. Now, it moves with the display.

image

### DIFF
--- a/src/views/2_SessionComponent.tsx
+++ b/src/views/2_SessionComponent.tsx
@@ -27,8 +27,7 @@ export class SessionComponent extends React.Component<Props, {}> {
     componentDidMount() {
         this.session.onscroll = () => {
             const shutterOffset = this.session.scrollTop;
-            this.shutter.style.top = `${shutterOffset}px`;
-            this.shutter.style.bottom = `-${shutterOffset}px`;
+            this.shutter.style.transform = `translateY(${shutterOffset}px)`;
         };
         this.props.session
             .on("job", () => this.props.updateStatusBar && this.props.updateStatusBar())
--- a/src/views/2_SessionComponent.tsx
+++ b/src/views/2_SessionComponent.tsx
@@ -26,7 +26,7 @@ export class SessionComponent extends React.Component<Props, {}> {
 
     componentDidMount() {
         this.session.onscroll = () => {
-            const shutterOffset = this.session.scrollTop
+            const shutterOffset = this.session.scrollTop;
             this.shutter.style.top = `${shutterOffset}px`;
             this.shutter.style.bottom = `-${shutterOffset}px`;
         };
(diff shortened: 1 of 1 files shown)


src/views/2_SessionComponent.tsx
```diff
@@ -25,6 +25,11 @@ export class SessionComponent extends React.Component<Props, {}> {
     }
 
     componentDidMount() {
+        this.session.onscroll = () => {
+            const shutterOffset = this.session.scrollTop;
+            this.shutter.style.top = `${shutterOffset}px`;
+            this.shutter.style.bottom = `-${shutterOffset}px`;
+        };
```
RobertoUa: it's better to user `transform: translate` if use update position frequently.

drew-gross:
Done.

src/views/2_SessionComponent.tsx
```diff
@@ -25,6 +25,10 @@ export class SessionComponent extends React.Component<Props, {}> {
     }
 
     componentDidMount() {
+        this.session.onscroll = () => {
+            const shutterOffset = this.session.scrollTop;
+            this.shutter.style.transform = `translateY(${shutterOffset}px)`;
+        };
```
shockone: I'm pretty sure this can be fixed used CSS only, which would be much faster.

drew-gross: I tried that for a long time, I'm pretty sure it's not possible with CSS only. The issue is that if the shutter div is on top of the content div, it captures the scroll events, and the content div can't be scrolled. There is no way to force the shutter div to be the same size as the content either. Stack Overflow couldn't come up with anything either: http://stackoverflow.com/questions/38760571/cover-a-div-with-a-transparent-overlay

shockone:
```
<div>
  <shutter/>
  <scrollable/>
</div/
```

If you have a structure like this, it wouldn't need to be scrolled at all. As far as I remember, it means we have to move the shutter one level up.

By the way, is there a better word for this thing than "shutter"?

shockone: I remembered that it was implemented already, but because of lack of tests I broke it later somehow. Please refer to this discussion and also take a look at where `jobsContainer` is used in that PR. https://github.com/shockone/black-screen/pull/580/files#r70489211

drew-gross: I tried implementing a few things, but none of them worked. This is how that old PR does it:

https://jsfiddle.net/7rvxpoaj/1/

But it doesn't work. I also tried some other stuff and couldn't make it work.

shockone: Fixed it in this commit. Closing. https://github.com/shockone/black-screen/commit/862ae28ee1204b12ac91a11ecf35eb15e9aedb95

PR #677, opened 2016-08-08
[WIP] Man page autocomplete tests

drew-gross:
I also had to move some code around due to some weird module loading behaviour I was seeing.


drew-gross:
This needs more work. Will reopen later.

PR #679, opened 2016-08-08
Grep auto complete

mkenne11:
I added additional auto-complete options to the grep auto-completion provider. Three types of options were added:

- Standard options with free-text values e.g. --version, --regexp.
- Options with a fixed set of values/ enumerated types e.g. --binary-files, --devices.
- Options who's values are file or directory paths e.g. --file, --exclude-dir.

### DIFF
--- a/src/plugins/autocompletion_providers/Grep.ts
+++ b/src/plugins/autocompletion_providers/Grep.ts
@@ -1,9 +1,375 @@
 import {PluginManager} from "../../PluginManager";
-import {longAndShortFlag, mapSuggestions} from "../autocompletion_utils/Common";
+import {longFlag, longAndShortFlag, mapSuggestions, anyFilesSuggestionsProvider,
+    Suggestion, styles, anyFilesSuggestions, directoriesSuggestions} from "../autocompletion_utils/Common";
 import combine from "../autocompletion_utils/Combine";
+import {AutocompletionContext} from "../../Interfaces";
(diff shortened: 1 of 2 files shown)


mkenne11:
Am keen to hear your thoughts on whether some of the auto-complete functions can be made more generic and added to the "autocompletion_providers/Common.ts" file.

The long flag options with file or directory path values seem likely candidates i.e. --file, --exclude-dir.

shockone:
Could you rebase the branch, please?

src/plugins/autocompletion_providers/Grep.ts
```diff
@@ -1,8 +1,278 @@
...
+import {AutocompletionContext} from "../../Interfaces";
+import {isDirectory} from "../../utils/Common";
 
-const options = combine([
-    mapSuggestions(longAndShortFlag("count"), suggestion => suggestion.withDescription("Only a count of selected lines is written to standard output.")),
+// Grep option suggestions based on linux  man file:
+// http://linux.die.net/man/1/grep
+const baseOptions = combine([
+    // Generic Program Information
+    mapSuggestions(longAndShortFlag("version", "V"), suggestion => suggestion.withDescription(`
```
shockone: Having a clear pattern it's better to create a data-only array of objects, and then map it to suggestions.

mkenne11:
I tried making the changes requested. Please let me know if any further changes are needed.

shockone:
It looks good.

PR #684, opened 2016-08-09
Man page autocomplete tests

drew-gross:
I also had to move some code around due to some weird module loading behaviour I was seeing.


src/plugins/autocompletion_utils/Combine.ts
```diff
@@ -0,0 +1,9 @@
+import * as _ from "lodash";
+import {AutocompletionContext, AutocompletionProvider} from "../../Interfaces";
+import {Suggestion} from "./Common";
+
+export default (providers: AutocompletionProvider[]): AutocompletionProvider => {
```
shockone: Default exports are bad. https://basarat.gitbooks.io/typescript/content/docs/tips/defaultIsBad.html

PR #689, opened 2016-08-12
Add ability to parse man pages with line breaks in a flag description, and cp autocomplete

drew-gross:
After this one I will make a PR that auto-detects the number of spaces in a line that is part of the same paragraph.


shockone:
👍

PR #692, opened 2016-08-12
Fix scanner for unicode characters

### LINKED ISSUE #675: Can't cd to directories with 'é' in their name
Hi,
It seems I need to add " for black-screen to cd to directories with an 'é' in it.
I also have an error you can see on the last screenshot each time I start it but I don't know if it's related to the problem: ```Uncaught (in promise) Error: ENOENT: no such file or directory, realpath '/home/terence/.rvm/rubies/default'(…)```

black-screen

black-screen2
 
*Arch Linux 4.6.4-1
black-screen 0.2.17-1 installed from the AUR using pacaur*

drew-gross:
Fixes #675

### DIFF
--- a/src/shell/Scanner.ts
+++ b/src/shell/Scanner.ts
@@ -177,7 +177,7 @@ const patterns = [
         tokenConstructor : SingleQuotedStringLiteral,
     },
     {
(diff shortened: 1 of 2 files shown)


shockone:
👍

terencode:
:+1:

PR #696, opened 2016-08-14
Improve git autocomplete

drew-gross:
Add options for `git commit`, and complete branch names for `git branch`


src/plugins/autocompletion_providers/Git.ts
```diff
@@ -34,11 +35,161 @@ function canonizeBranchAlias(alias: string) {
...
+    },
+    {
+        longFlag: "no-status",
+        description: "Do not include the output of git-status(1) in the commit message template when using an editor to prepare the default commit message.",
+        noShortFlag: true,
+    },
+    {
+        longFlag: "no-gpg-sign",
+        description: "Countermand commit.gpgSign configuration variable that is set to force each and every commit to be signed.",
+        noShortFlag: true,
```
shockone: Can we determine it based on `shortFlag` key presence?

drew-gross: That has a separate meaning

key present -> use it
key not present -> use the first letter of the long flag
noShortFlag: true -> there is no shortFlag

Issue #577, opened 2016-07-08
[cmd + k] Clear screen with a shortcut


darol100:
### Problem / Motivation

In Mac there is the ability of clear the screen by using the `cmd + k`.  In ubuntu, the shortcut is `ctrl + l`.
### Solution

Add the ability of clearing the screen with `cmd + k`. This shortcut is the same as the command `clear`.

ChrisButterworth:
`ctrl + l` clears terminal and works in black screen

luco:
@ChrisButterworth Yes! But I think everyone is used to `cmd+k`

admix:
@luco ```cmd+k``` is already used by Pane->Previous.. which is not that great imo.. mb combinations like: ```cmd+]``` and ```cmd+[``` would be better, however its a topic to the other issue.

dperrymorrow:
i was confused by cmd+k not working at first as well, then i discovered that clear is a suggested command
![black screen 2017-05-18 15-16-06](https://cloud.githubusercontent.com/assets/64978/26225637/fed4fecc-3bdc-11e7-806d-22bc8755d21f.png)

vlad-shatskyi:
Closing. Please use <kbd>Ctrl</kbd>+<kbd>L</kbd>

PR #718, opened 2016-08-29
Added Keybinding support & Menu support for Linux/Windows

Islandman93:
Allows editing of default keybindings by editing src/views/keyevents/Keybindings.ts
Adds menu support for Windows & Linux by using the 'CmdOrCtrl' accelerator, by default copy is bound to Ctrl+Shift+C for Windows/Linux.
All the keybinding defaults were set from the existing code.

### DIFF
--- a/src/Enums.ts
+++ b/src/Enums.ts
@@ -97,3 +97,47 @@ export enum SplitDirection {
     Vertical,
     Horizontal,
 }
+
+export enum KeyboardAction {
+    // CLI commands
+    cliRunCommand,
+    cliInterrupt,
+    cliClearJobs,
+    cliDeleteWord,
+    cliClearText,
+    cliAppendLastArgumentOfPreviousCommand,
+    cliHistoryPrevious,
+    cliHistoryNext,
+    // autocomplete commands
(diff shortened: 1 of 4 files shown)


Islandman93:
Tested on Ubuntu 14.04 all keybindings and menu actions work. Windows support is untested, but should work. Mac should remain 100% unchanged, but I don't have a mac to test.
npm run test returned no errors. 

I didn't write any tests for my code because I think it's unneeded, it's just simple array searching/filtering.

In the future it would be great if all of this could reside in a cson object. For now it resides as code in Keybindings.ts because it was easier. But with this setup it could be converted into loading from cson without changing the UserEventHandler.ts. 

Fixes #577 
Fixes #441

shockone:
Thanks, @Islandman93. I've fixed two typos and used `find` instead of `filter` here: https://github.com/shockone/black-screen/commit/45e5930a8f6ac488f005738c1319509ec560d086

Islandman93:
Wow typos in both functions... No more late night pull requests for me.

Thanks for fixing and merging!

PR #731, opened 2016-09-02
Break plugin dependency on PluginManager

drew-gross:
Currently all of the plugins import `PluginManager`. This works for now, but it we people to be able to eventually add plugins that live outside this repo, it won't work so great. This breaks the dependency of the `preexec` plugins (currently only the one) on `PluginManager`. If you think this approach is good I'll do the same for the other 3 plugin types.


shockone:
I was thinking about it as well. The more general question is how to compile your plugins against BlackScreen without requiring BlackScreen. I'm not sure I understand how the other import

```typescript
import {Job} from "../../shell/Job";
```

will work in this implementation. 

As an investigation path, I know that VSCode also uses TS and supports plugins. Maybe we could use the same approach as they do.

drew-gross:
The `Job` import is just used for the type, so it could be replaced with `any` or the file could be converted to plain JS to break that dependency. Atom is written in coffeescript, but plugins are written in JS, or some compile-to-js language that is compiled before submitting as a package. I think VSCode is the same way, they encourage you to use TS, but you still have to compile the TS before VSCode can load it. I think it would probably have to be the same for black-screen. If someone is building a plugin in TS we can provide typing if the want to "build against" black-screen, but that would be optional if you just wanted plain JS.

shockone:
I'd still like us to take a closer look at VSCode. Other projects have the same issue too: Intellij IDEA, Eclipse, etc.

drew-gross:
Here are the docs for how to get started with an Eclipse plugin. It looks like a similar setup, just way more heavyweight: https://www.eclipse.org/articles/Article-Your%20First%20Plug-in/YourFirstPlugin.html

It ends up loading a jar file, so you could build in Scala or any other JVM language. Eclipse also provides "Extension points" for plugins, and an interface you are supposed to implement, so it seems pretty similar to what we've got going on here, except adapted to Java. 

IntelliJ also seems somewhat similar: http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started.html

shockone:
Let's delay this decision until after we have a real need to support third-party plugins. We'll have more information then, and maybe there'll appear some standard way of doing this in the community.

vshatskyi:
@drew-gross I plan to close old PRs to keeps things clean. Is it OK with you if I close this one?

PR #747, opened 2016-09-12
[wip] LS decorator

drew-gross:
This is a decorator for LS that essentially turns the text of each item into a link that opens the item in your default editor.

ls-decorator

I wanted to get some feedback on what important features people might want in this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "@types/pty.js": "0.2.30",
     "@types/react": "0.14.33",
     "child-process-promise": "2.1.3",
+    "dirStat": "0.0.2",
     "font-awesome": "4.6.3",
     "fs-extra": "0.30.0",
     "fuzzaldrin": "2.1.0",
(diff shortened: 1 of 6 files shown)


shockone:
That's awesome. I have a few ideas:
* I think it would be helpful to add an underline on hover to indicate its clickable.
* Can we generalize it work on output of every command (where only output words that are valid paths would be clickable) while still limiting it to `ls` for now?

drew-gross:
Underline on hover would be awesome, but it's difficult to style pseudo-elements with the inline styles this project uses. In other projects I've used webpack+css modules, which worked well, I could set that up for this project if you want. Or I could add it as a global style, but I think thats pretty messy. Working on the output of every command would be nice but I'm not sure how to accomplish it. iTerm 2 tries to do that, but it doesn't do very well on files/links that have spaces or certain other non-alpha characters in them.

drew-gross:
Ok I implemented underline on hover in a bit of a hacky way. I looked into switching to webpack, but it will be tricky, so I'm just going to merge this now. Generalizing to work on all output is something I'd like to work on later, as long as it actually works on all output, instead of most output like iTerm 2 does.

PR #748, opened 2016-09-12
[wip] refresh git status after command

mathewmeconry:
Hey

I made some more changes to to the git status bar.
It has now an indicator for uncommited changes and should refresh after each command to update the bar if there is any change in the repo.

Unfortunately I'm unable to apply the refresh with the afterExecPlugin. 

Can you help me there?

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "black-screen",
   "productName": "Black Screen",
   "description": "A terminal emulator for the 21st century.",
-  "version": "0.2.32",
+  "version": "0.2.33",
   "main": "compiled/src/main/Main.js",
   "author": "Volodymyr Shatskyi <shockone89@gmail.com>",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "black-screen",
   "productName": "Black Screen",
   "description": "A terminal emulator for the 21st century.",
-  "version": "0.2.31",
+  "version": "0.2.32",
   "main": "compiled/src/main/Main.js",
   "author": "Volodymyr Shatskyi <shockone89@gmail.com>",
   "repository": {
--- a/src/Interfaces.ts
+++ b/src/Interfaces.ts
@@ -61,6 +61,10 @@ export interface PreexecPlugin {
     (job: Job): Promise<void>;
 }
 
+export interface AfterexecPlugin {
+    (job: Job): Promise<void>;
+}
+
 export interface TerminalLikeDevice {
     screenBuffer: ScreenBuffer;
     dimensions: Dimensions;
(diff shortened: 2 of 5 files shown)


shockone:
I'm not sure what you're trying to achieve with the after exec plugin. Binding refreshes to after exec isn't the best way to go: if I `git push` from my iTerm, the status bar should be updated immediately, even though I executed push from a different process.

mathewmeconry:
good input. I try it

mathewmeconry:
do you think this is the better way?

image

shockone:
I think events are a much superior way, even though we have to write more code for them. With intervals you blindly try checking the status every so often, because you don't have enough information to make a better guess, whereas events provide this additional knowledge you need to perform an action only when it's required.

mathewmeconry:
After a short search I can't find a eventhandler for git changes but I'll do some research

shockone:
Again, I'm not sure what changes you want to watch, but I bet `.git/refs/remotes/origin/HEAD` changes when you push. You can try watching the whole `.git/refs/remotes/` directory, but make sure it doesn't cause CPU to go crazy.

mathewmeconry:
I'd like to watch following changes:
Uncommited files
Branch checkouts
Unpulled changes
Unpushed changes

So it would be the cheapest way to update these stats with a intervall

mathewmeconry:
When I look into the git implementation of vscode it looks like, they use some kind of infinite loop and listen on local file changes. Which I think is an overkill for a terminal where the git implementation is only a nice feature.

https://github.com/Microsoft/vscode/blob/96afb61cbfff286c89c6634267a9bb7acadf3fb2/src/vs/workbench/parts/git/browser/gitServices.ts

shockone:
Uncommited files and Branch checkouts should work as it is now, don't they?

mathewmeconry:
Unfortunately the git status only gets updated on directory changes in the current master

shockone:
Oh, well, then it must have been broken recently. It used to work in real time.

mathewmeconry:
I only tested it on Fedora. I don't know if it works on OS X.

Do you wanna fix this or should I do some last tests with this pull request?

shockone:
I'm afraid I won't be able to fix it any time soon, it's been a busy month.

mathewmeconry:
ok. shall I finish this pull request with the intervall update?

shockone:
Sure, if it's hard to do with events.

mathewmeconry:
I'll finish it this week. hopefully I find any good solution to do it with events

shockone:
In my understanding, the current solution should be suitable to track what you want. You'd just need to change the ignored pattern here https://github.com/mathewmeconry/black-screen/blob/dcfcd9b596754827307adfeae22874890646ae0f/src/plugins/GitWatcher.ts#L43. The question is why it doesn't get updated in real time right now.

vshatskyi:
@mathewmeconry I plan to close old PRs to keeps things clean. Is it OK with you if I close this one?

mathewmeconry:
@vshatskyi Sure. Sorry I had no time left to finish this pull

PR #755, opened 2016-09-17
Added ps auto-complete functionality

mkenne11:
This PR provides auto-complete functionality for the "ps" command and some of the the options available.

The long-flag options provide suggested values based on the values available. These values can be chained together separated by commas. The example in the screenshot attached is for the "--pid" (process id) option and shows a number of values entered separated by commas. 
image

I would like to cache calls for long-flag option to save multiple calls to fetch available options. In the "src/utils/Process.ts" file I attempted to do this by using memoize to cache the results of the processes function (see annotation inline). Unfortunately the result is cached for the entire terminal session - I need to cache this for each terminal command instance (if that makes sense).

Please let me know if you can think of any ways of caching per command instance.

### DIFF
--- a/src/utils/Process.ts
+++ b/src/utils/Process.ts
@@ -1,0 +1,85 @@
+import {executeCommandWithShellConfig} from "../PTY";
+import {intersection} from "lodash";
+
+// http://linuxcommand.org/man_pages/ps1.html
+
+export interface User {
+    ruserid: string;
+    ruser: string;
+    euserid: string;
+    euser: string;
+}
+
+export interface Group {
+    rgroupid: string;
+    rgroup: string;
+    egroupid: string;
+    egroup: string;
+}
+
+export interface Terminal {
+    name: string;
+    ruser: string;
+}
+
+export interface Process {
+    pid: string;
+    time: string;
+    ruser: string;
+    cmd: string;
+}
+
+export interface Session {
+    sid: string;
+    ruser: string;
+    rgroup: string;
+}
+
+const ignoreUsers: string[] = ["nobody"];
+
+const ignoreGroups: string[] = ["nobody"];
+
+const resultSet = (result: string[]) => {
+        const numColumns = result[0].trim().replace(/ +(?= )/g, "").split(" ").length;
+        return result.splice(1)
+                     .map(i => i.trim().replace(/ +(?= )/g, "").split(" ", numColumns));
+    };
+
+export const users =
+    async(): Promise<User[]> => {
+        const pInfo: string[][] =
+                resultSet(await executeCommandWithShellConfig("ps -eo ruid,ruser,euid,euser"));
+        return pInfo.map(p => <User> {ruserid: p[0], ruser: p[1], euserid: p[2], euser: p[3]})
+                    .filter(p => intersection(ignoreUsers, [p.ruser, p.euser]).length === 0);
+    };
+
+export const groups =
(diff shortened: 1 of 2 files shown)


src/utils/Process.ts
```diff
@@ -0,0 +1,102 @@
...
+export const processes =
+    async(): Promise<Process[]> => {
+        const pInfo: string[] = await executeCommandWithShellConfig("ps -eo pid,time,ruser,cmd");
+        return pInfo.splice(1)
+                    .map(p => {
+                        const v = p.trim().replace(/ +(?= )/g, "").split(" ", 4);
+                        return <Process> {pid: v[0], time: v[1], ruser: v[2], cmd: v[3]};
+                    });
+    };
+/*
```
mkenne11: Attempted to cache results of "processes" function for each command instance. Unfortunately this caches results for entire terminal session :(

src/plugins/autocompletion_providers/Process.ts
```diff
@@ -0,0 +1,188 @@
+import {PluginManager} from "../../PluginManager";
```
drew-gross: Our convention is to use the name of the command as the name of the file. Please changes this file to `Ps.ts`

src/plugins/autocompletion_providers/Process.ts
```diff
@@ -0,0 +1,188 @@
...
+                            style: styles.optionValue}));
+    };
+
+const processSuggestions = async(context: AutocompletionContext): Promise<Suggestion[]> => {
+        const arg = argInfo(context);
+        const processes = await Process.processes();
+        return processes
+                .filter(i => !arg.params.includes(i.pid))
+                .map(i => new Suggestion({value: arg.start + i.pid, displayValue: i.pid,
+                            description: "Process with ID '" + i.pid +
```
drew-gross: Please use backticks style strings, like `` `Process with ID ${i.pid}...` `` for these strings. Also the information shown on the left portion of the autocompletion (the actual PID) probably doesn't need to be shown on the right as well. The command is probably the most important thing to show in the details, and the user the second most important thing. See if you can shrink down this text to make sure there is aways space for the command (and hopefully user as well) here:

image

Similarly for the group, etc.

src/utils/Process.ts
```diff
@@ -0,0 +1,102 @@
...
+const ignoreUsers: string[] = ["nobody"];
+
+const ignoreGroups: string[] = ["nobody"];
+
+export const users =
+    async(): Promise<User[]> => {
+        const pInfo: string[] = await executeCommandWithShellConfig("ps -eo ruid,ruser,euid,euser");
+        return pInfo.splice(1)
+                    .map(p => {
+                        const v = p.trim().replace(/ +(?= )/g, "").split(" ");
```
drew-gross: `p.trim().replace(/ +(?= )/g, "").split(" ")` is repeated in a bunch of places, can you extract that into a function?

src/utils/Process.ts
drew-gross: Don't worry about it too much. `ps` is a fast command.

mkenne11:
No worries Drew the changes look straight forward - I'll get onto to them.

mkenne11:
@drew-gross I attempted to make the changes suggested. Please let me know if anything else needs to be fixed.

drew-gross:
Yep that looks good. Thanks for the help!

PR #762, opened 2016-09-20
Tab navigation

noygal:
Hi, you have an awesome app, just noticed it couple of days ago and no I'm hooked!

One thing I was missing was the tab navigation (prev/next) functionality, I took the liberty to implement it, I tried to follow the code style and tested it manually on MacOS 10.11.6. You did a great work and the source code is very readable, so it took couple of hours to implement the feature, the development environment worked without glitches. I would recommend to switch to stateless components design model though, as it seem like the app has high view-logic dependencies.

If your looking for collaborators, I'm up for the task, I have experience with nodejs, electron, react and typescript.

P.S
I also bound the meta key to the indexed tab switch (keeping the old ctrl bind in tact), the tab header place the meta key icon next to the tab index so UX wise the user expect to use it for the index tab switching (I had to peek at the source code to find the ctrl+index binding), apart from that MacOS bind ctrl+index to workspace switching so switching to the first tab was impossible for me.

### DIFF
--- a/src/views/keyevents/Keybindings.ts
+++ b/src/views/keyevents/Keybindings.ts
@@ -124,6 +124,18 @@ export const KeybindingsForMenu: KeybindingMenuType[] = [
         accelerator: "CmdOrCtrl+T",
     },
     {
+        action: KeyboardAction.tabPrevious,
+        accelerator: "Ctrl+Shift+Tab",
+    },
+    {
+        action: KeyboardAction.tabNext,
+        accelerator: "Ctrl+Tab",
+    },
+    {
+        action: KeyboardAction.tabClose,
+        accelerator: "CmdOrCtrl+Shift++W",
+    },
+    {
         action: KeyboardAction.panePrevious,
         accelerator: "CmdOrCtrl+K",
     },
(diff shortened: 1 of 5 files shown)


src/views/keyevents/Keybindings.ts
```diff
@@ -124,14 +125,26 @@ export const KeybindingsForMenu: KeybindingMenuType[] = [
...
+        accelerator: "Ctrl+Shift+Tab",
     },
     {
         action: KeyboardAction.tabNext,
-        accelerator: "CmdOrCtrl+J",
+        accelerator: "Ctrl+Tab",
     },
     {
         action: KeyboardAction.tabClose,
+        accelerator: "CmdOrCtrl+Shift++W",
```
drew-gross: The standard in Mac apps seems to be Cmd+Shift+W closes a window, not a tab. (try it in Chrome, Sublime, iTerm, etc.) `black-screen` has both tabs and panes so it's a bit more complicated, but I still think Cmd+Shift+W should be reserved for closing a window.

noygal: I didn't understood why the Cmd+Tab was hooked to the close pane event in the first place :)

I changed it to the close tab to Cmd+W, and the close pane to Cmd+P - it close to the J/K pane navigation and it usually reserve to send to printer on mac apps, so not much usage on terminal app.

drew-gross:
Awesome, looks good now. I will merge as soon as tests pass.

PR #785, opened 2016-09-27
Revert all changes since "update deps", and make tests run a build before passing

drew-gross:


drew-gross:
Tests passed, which should (now) mean that the deploy will be successful. Hopefully.

PR #828, opened 2016-10-15
Add missing vt100 events

noygal:
I had some problems with the backspace operation on ssh so I took a crack and found some unsupported events. I added the missing event and the relevant docs links.

I also fixed: https://github.com/shockone/black-screen/issues/688
For some reason the docs didn't mention `3` param...

Tested on macOS sierra, ssh to debian jessie.

Also fixed two small react warnings.


noygal:
I also see that working on master preserve previous commit on PR, I'll refork  and branch to prevent that in the future :)

drew-gross:
Looks awesome, thanks for the help! I usually squash merge so reforking isn't necessary.

PR #836, opened 2016-10-20
AppImage for Linux

probonopd:
Please do **not** merge yet, running into seemingly unrelated issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,24 @@
 osx_image: xcode7.3
 language: node_js
+
 os:
   - osx
   - linux
+  
 cache:
   directories:
     - node_modules
+    
+script:
+  - npm run update-dependencies
+  
 deploy:
   skip_cleanup: true
   provider: script
   script: bash housekeeping/deploy.sh
   on:
     branch: master
+    
 env:
   global:
     # GH_TOKEN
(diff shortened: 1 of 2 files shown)


drew-gross:
Looks like a build error. You may have to massage some dependencies. Also, this PR as is removes the OSX build. We'll need both OSX and linux builds working before we can merge this.

probonopd:
Reinstated macOS, have no clue about the dependencies; this doesn't seem to be AppImage specific to me and I don't know how I can solve these.

drew-gross:
You can use `npm run update-dependencies` to update all dependencies to the latest version. If that doesn't work, then we'll probably to put in some real effort.

probonopd:
Ping @drew-gross

drew-gross:
This PR has some merge conflicts and also looks like it still isn't passing tests.

.travis.yml
```diff
@@ -1,16 +1,24 @@
...
+  - osx
+  - linux
+  
 cache:
   directories:
-  - node_modules
+    - node_modules
+    
+script:
+  - npm run update-dependencies
```
shockone: We don't want to have different dependency versions than in `package.json`. It can break things.

probonopd: I think it should be fine to just use the latest `electron-builder` and remove the `npm run update-dependencies`

vshatskyi:
@probonopd I plan to close old PRs to keeps things clean. Is it OK with you if I close this one?

probonopd:
Yes

PR #838, opened 2016-10-21
Git status plugin

drew-gross:


drew-gross:
Add some enhancements to Git Status. It's structure fairly well I think so further enhancements should be pretty easy hopefully.

gitstatusplugin

Issue #715, opened 2016-08-26
git-prompt: current diff status


dmccaffery:
This is my usual prompt:

<img width="536" alt="screen shot 2016-08-25 at 11 25 16 pm" src="https://cloud.githubusercontent.com/assets/1803684/17996042/33df4656-6b1b-11e6-907c-12cf54665e21.png">

black-screen is gorgeous, and I am definitely loving it, but it does hide the default prompt. While it shows the current branch in the bottom right corner, it does not show the current state (diff) like my prompt does.

If you would be able to include the +adds, -deletions, and ~changes somewhere near the git branch label, I could switch to black-screen full-time.

One more question: does black-screen use fira code as the font? I cat'd a file and saw the ligatures. If so... thats just awesome sauce.

For more information on my prompt: https://github.com/pulsebridge/uber

dmccaffery:
I'll look at contributing to this shortly... is there a design doc / approval process?

drew-gross:
No "official" approval process. This looks like a good idea, I'm a little afraid of the status bar getting to busy though.

wy193777:
Just created pull request #840 that implemented this function.

PR #840, opened 2016-10-23
Implemented Issue 715's function

wy193777:
- status bar will show file change status
- fix a bug all git directory treated as dirty

The implementation of issue #715.

### DIFF
--- a/src/plugins/GitWatcher.ts
+++ b/src/plugins/GitWatcher.ts
@@ -102,42 +102,38 @@ class GitWatcher extends EventEmitter {
         });
     }
 }
-
 function linesToFileChanges(lines: string): FileChanges {
-
-  let fileChanges = {
-    stagedAdded: 0,
-    stagedModified: 0,
-    stagedDeleted: 0,
-    stagedUnmerged: 0,
-
-    unstagedAdded: 0,
-    unstagedModified: 0,
-    unstagedDeleted: 0,
-    unstagedUnmerged: 0,
-  };
+  let stagedChanges = new Map<string, number>([["+", 0], ["~", 0], ["-", 0], ["!", 0]]);
+  let unstagedChanges = new Map<string, number>([["+", 0], ["~", 0], ["-", 0], ["!", 0]]);
   lines.split("\n").slice(1).forEach((line) => {
     switch (line[0]) {
-      case "A": fileChanges.stagedAdded += 1; break;
+      case "A": stagedChanges.set("+", stagedChanges.get("+") + 1); break;
       case "M":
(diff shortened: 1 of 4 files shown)


drew-gross:
Awesome, can you include some screenshots too show what it looks like?

wy193777:
screenshot from 2016-10-23 21-57-25

Staged file changes will be shown in green and unstaged will be shown in red. 
Numbers after '+', '~', '-' mean added, changed and deleted files. 

Just pushed a new commit fixed a bug that unstaged change will be counted as staged.

drew-gross:
Looks pretty good. How would you feel about hiding the ones the are zero? I want to avoid the status bar getting too cluttered.

wy193777:
Those numbers will not show if the repository is clean or not a git repository. 

 Yes, I think that's a good idea. I will try to finish this on this week.

wy193777:
Just updated. It will not show 0s now.

drew-gross:
Awesome, thanks. I removed a `console.log` for you and I will merge when tests pass. BTW we have some git related utilities here: https://github.com/shockone/black-screen/blob/master/src/utils/Git.ts that you may find useful in the future. I kinda like your API better though, maybe we should adopt that one throughout the codebase.

dmccaffery:
I was looking into this too! You 'da man, @wy193777

dmccaffery:
... or woman, whichever the case may be... either way, I <3 you!

dmccaffery:
screen shot 2016-10-31 at 1 11 23 am

The only thing I don't like is the mapping between the local branch and the remote branch... that definitely makes it seem cluttered.

drew-gross:
Yeah, I know. It is somewhat useful, but also very cluttered. Maybe we can have that show when you hover your mouse, and when you change branches, but otherwise slide away.

wy193777:
The status on bottom is a little bit verbose after we have file change number.

dmccaffery:
@drew-gross @wy193777 Don't get me wrong; I LOVE this feature, which is why I opened #715 to begin with... I like the simple idea of clicking on the area to "show/hide" additional detail... maybe a three click rotation as follows:

1) branch name only (like it was before)
2) @wy193777 changes with simple branch (no remote map)
3) full detail like it is now

Just keep clicking to rotate through and persist the state

drew-gross:
That sounds like a pretty solid plan. It's kinda like a setting, but also kinda like a persistent UI feature.

wy193777:
Maybe create a new issue about this?

drew-gross:
Good idea. Opened in #862

PR #852, opened 2016-10-31
Fix alt key

drew-gross:
I'm not 100% sure about this fix. It causes `sleep 10` then `alt+d` to display the same thing as in `iTerm` when in "Alt key = ESC+" mode. I was never able to make vim work the way @Alok described in either `black-screen` or `iTerm`. @shockone would you be able to review this one? I'm not very familiar with the ANSI/escape codes stuff.

### DIFF
--- a/src/shell/Job.ts
+++ b/src/shell/Job.ts
@@ -113,7 +113,19 @@ export class Job extends EmitterWithUniqueID implements TerminalLikeDevice {
         if (typeof input === "string") {
             text = input;
         } else {
-            text = input.ctrlKey ? String.fromCharCode(input.keyCode - 64) : normalizeKey(input.key, this.screenBuffer.cursorKeysMode);
+            if (input.ctrlKey) {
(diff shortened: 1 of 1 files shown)


src/shell/Job.ts
```diff
@@ -113,7 +113,19 @@ export class Job extends EmitterWithUniqueID implements TerminalLikeDevice {
         if (typeof input === "string") {
             text = input;
         } else {
-            text = input.ctrlKey ? String.fromCharCode(input.keyCode - 64) : normalizeKey(input.key, this.screenBuffer.cursorKeysMode);
+            if (input.ctrlKey) {
+                text = String.fromCharCode(input.keyCode - 64);
+            } else if (input.altKey) {
+                let char = String.fromCharCode(input.keyCode);
```
shockone: Isn't `input.key` already in the proper case?

drew-gross: From my testing `input.key` is the special char macOS generates, ∂ for alt+d for example.

Issue #873, opened 2016-11-06
Better Autocomplete Search


pmkary:
Autocomplete in BS only matches the start of the identifiers which is a bit bad. Good autocomplete search finds the given string as a sequence of characters and then finds it as a sequence in identifiers, such as:

<img width="548" alt="screen shot 2016-11-06 at 11 17 05 pm" src="https://cloud.githubusercontent.com/assets/2157285/20041017/89df015a-a477-11e6-9d6e-de25d599a4c0.png">

I once tried to make an suggestion provider for my language and then I tried to make [an editor for it, which is available here for testing](http://www.karyfoundation.org/research/playgrounds/gerda/02/). And it gives you something like this:

<img width="256" alt="screen shot 2016-11-06 at 11 07 53 pm" src="https://cloud.githubusercontent.com/assets/2157285/20040931/0658f0bc-a476-11e6-9283-dabf855138ae.png">

I said that because I have an implementation for this algorithm in TypeScript, so you can use it with some changes. It also inserts spans in the text to make it color (what chars are in the sequence):

```TypeScript
export function FilterSuggestions( suggestions: Array<string> , userInput: string ) {
    var results = new Array<string>( );
    for ( var index = 0; index < suggestions.length; index++ ) {
        var element = suggestions[ index ];
        var currentSearchCharIndex = 0;
        var elementToBeAdded = '';
        for ( var searchStringsIndex = 0; searchStringsIndex < element.length; searchStringsIndex++) {
            var currentChar = element[ searchStringsIndex ];
            if ( currentChar == userInput[ currentSearchCharIndex ] ) {
                if ( currentSearchCharIndex < userInput.length ) {
                    elementToBeAdded += '<span class="GerdaItemHighlighters">' + currentChar + '</span>';
                    currentSearchCharIndex++;
                } else {
                    elementToBeAdded += currentChar;
                }
            } else {
                elementToBeAdded += currentChar;
            }
        }
        if ( currentSearchCharIndex >= userInput.length ) {
            results.push( elementToBeAdded );
        }
    }
    return results;
}
```

I tried to implement it to the BS but the code base is a bit strange and I don't get where to start. So if you give me some points on where to implement these stuff I can do this myself.

drew-gross:
Cool. Black Screen uses React so you will have to modify your algorithm somewhat, as appending spans as text isn't really how React works.

[...]

PR #877, opened 2016-11-08
Autocomplete Sequence Search

pmkary:
So we are all aware of #873. Let's have the conversation on "ideas" there, "implementation" here


src/views/AutocompleteComponent.tsx
```diff
@@ -1,21 +1,25 @@
...
+const SuggestionComponent = ({suggestion, onHover, onClick, isHighlighted, searchKey}: SuggestionProps) =>
     <li style={css.autocomplete.item(isHighlighted)}
         onMouseOver={onHover}
         onClick={onClick}>
 
         <i style={Object.assign({}, css.suggestionIcon, suggestion.style.css)} dangerouslySetInnerHTML={{__html: suggestion.style.value}}/>
-        <span style={css.autocomplete.value}>{suggestion.displayValue}</span>
+
+        <span style={css.autocomplete.value}
+              dangerouslySetInnerHTML={{__html: HighlightSequencedSuggestion( suggestion.displayValue, searchKey)}} />
```
drew-gross: Please don't use `dangerouslySetInnerHTML` here. That is for things like markdown formatters that have their own xss protection built in, and where you don't have control over the underlying implementation. Since we are writing the `HighlightSequencedSuggestion`, we can have return a structure that can be converted to `React` elements, rather than a string of HTML.

src/main/Main.ts
```diff
@@ -42,7 +45,7 @@ app.on("ready", () => {
     app.on("open-file", (event, file) => browserWindow.webContents.send("change-working-directory", file));
 });
 
-app.on("mainWindow-all-closed", () => process.platform === "darwin" || app.quit());
+app.on("window-all-closed", app.quit );
```
drew-gross: Is this to fix that one issue about quitting when all windows are closed?

pmkary: yeah sorry. My commits are so strange... I can't really work on one topic at a time. But yes. For now simply ignore it, I will make the Black Screen work in multi windows, but for now it would just fix the crash when you close a window but your menu is still there and does not function...

src/views/AutocompleteComponent.tsx
pmkary: I thought that is better for speed reasons. But okay I'll make it react component based.

src/SequenceKit.ts
```diff
@@ -0,0 +1,53 @@
+// Author: Pouya Kary <k@karyfoundation.org> (in case you had any questions...)
+
+import {colors} from "./views/css/colors";
+import {Suggestion} from "./plugins/autocompletion_utils/Common";
+
+export interface SortableObject {
```
shockone: Not used.

src/SequenceKit.ts
```diff
@@ -0,0 +1,53 @@
...
+    weight: number;
+    value: Suggestion;
+}
+
+/**
+ * Checks if the __key__ exists as a _sequence_ in the __element__.
+ */
+export function SequenceFilter(element: string, key: string): boolean {
```
shockone: Please don't start function names with uppercase letters. It implies it's a constructor function.

shockone: Can we make this function return an array of indices of matched letters? An empty array would mean `false`, and we could reuse those indices to highlight letters instead of implementing the algorithm twice.

pmkary: I though of the same thing but wanted to change as much as less design as possible...

vshatskyi:
@pmkary I plan to close old PRs to keeps things clean. Is it OK with you if I close this one?

PR #903, opened 2016-11-19
Trim input before scanning

drew-gross:
The tests say the scanner should generate an `Invalid` token, but I don't see why they should do that. Esp. for plugins, they shouldn't care that the input had whitespace.

Tagging @shockone because this is a change to tested behaviour so I would appreciate a second set of eyes.


shockone:
The point was to have a parse tree isomorphic to the input string. With thins change if I parse three spaces I can't serialize it back to three spaces.

shockone:
Does it cause some issues, @drew-gross?

drew-gross:
Ah I see. For the plugins I'm writing I want to match against the parsed line to see if the command is applicable. Eg. for the `git branch` plugin I check if the parsed command line equals `[Word("git"), Word("branch")]`. But I also want the plugin to activate when the command line contains `"   git    branch   "`, so I made this change to ensure that `"git branch"` and `"    git    branch   "` parse to the same thing.

If the parse tree being isomorphic is important, I could come up with a different strategy.

vshatskyi:
@drew-gross should we close this? Also, I believe that currently the examples you mentioned do get parsed to the same thing, it's a string containing only spaces that differs.

drew-gross:
No, `"git branch "` definitely gets parsed to `[Word, Word, Invalid]` before this PR. I can update this PR to pass a modified parse tree to the plugins.

vshatskyi:
Oh, that doesn't sound right. Perhaps the right solution would be for a node to "eat" all the preceding *and* the following spaces instead of only the former ones.

drew-gross:
That seems like a good idea actually. I'll do that one.

vshatskyi:
One more thing: please check where `LeafNode.spaces()` is used, and change the logic there to use spaces from both ends. Probably we can rename this one to `precedingSpaces` and name the new method `followingSpaces`.

vshatskyi:
Oh, and `fullEnd` should be changed as well.

drew-gross:
@vshatskyi I added the `followingSpaces`/`preceedingSpaces` like you suggested but I'm not sure what should be changed about `fullEnd`.

vshatskyi:
You're right, it shouldn't be changed. Sorry about that.

vshatskyi:
@drew-gross it seems that `followoingSpaces` broke autocompletion after a space. Try typing `git `. Previously it'd show the next token suggestions.

drew-gross:
Whoops sorry :( I created a new issue.

PR #923, opened 2016-12-07
Enable Multi-Window Support

### LINKED ISSUE #365: Application quits when window closed on OSX
Great idea for an application! I noticed that when the window is closed on OSX, the application quits. As noted in the [electron quick start](http://electron.atom.io/docs/tutorial/quick-start/#write-your-first-electron-app), the expected behaviour is for the application to stay active until the user quits (cmd-q).

It appears you attempted to replicate the behaviour suggested in the quick start on [line 14](http://electron.atom.io/docs/tutorial/quick-start/#write-your-first-electron-app) but it's not quite working. My first guess is that it should be `window-all-closed` instead of `mainWindow-all-closed`, but I can't test it right now so just wanted to get the issue posted in the meantime. Happy to work on this if the fix isn't obvious.

### LINKED ISSUE #922: Multi-Window support?
Are there any plans on supporting multiple windows?

What would be the steps to implement it?

jsphpl:
- Allow multiple windows (fixes #922)
- Keep application running on osx after all windows have been closed (fixes #365)

Changes and refactoring applied to achieve this:
- Store window instances in `windows` global in `main/Main.ts`
- Introduce methods for window handling including focussed window in `main/Main.ts`
- Add menu items and keybindings for new/close window
- Make the `electronWindow` a property of the `ApplicationComponent` so we can `.close()` it when all tabs are gone
- Move `Menu.ts` to `src/main/` and register it from within the main process, don't bind the handlers to specifit windows
- Create handlers in `ApplicationComponent` for all menu events so we can call them on the currently focussed window
- Move `Keybindings.ts` to `src/`

### DIFF
--- a/src/main/Main.ts
+++ b/src/main/Main.ts
@@ -25,14 +25,14 @@ app.on("ready", () => {
     createWindow();
     registerApplicationMenu();
 
-    globalShortcut.register('CommandOrControl+N', () => {
+    globalShortcut.register("CommandOrControl+N", () => {
         createWindow();
-    })
+    });
 });
 
 app.on("window-all-closed", () => {
     // On OSX the app usually keeps running until Cmd+Q'd.
-    if (process.platform !== 'darwin') {
+    if (process.platform !== "darwin") {
         app.quit();
     }
 });
@@ -83,9 +83,9 @@ export function createWindow() {
         browserWindow.focus();
     });
 
-    browserWindow.on('closed', () => {
+    browserWindow.on("closed", () => {
         closeWindow(browserWindow);
-    })
+    });
 
     windows.push(browserWindow);
 }
@@ -103,8 +103,7 @@ function closeWindow(browserWindow: Electron.BrowserWindow) {
     let next_window = Math.min(index, windows.length);
     if (next_window > -1) {
         windows[next_window].focus();
-    }
-    else {
+    } else {
         focussedWindowIndex = -1;
     }
 }
--- a/src/main/Main.ts
+++ b/src/main/Main.ts
@@ -1,12 +1,52 @@
-import {app, ipcMain, nativeImage, BrowserWindow, screen} from "electron";
+import {app, nativeImage, BrowserWindow, screen, globalShortcut, Menu} from "electron";
 import {readFileSync} from "fs";
 import {windowBoundsFilePath} from "../utils/Common";
+import {buildMenuTemplate} from "./Menu";
+
+/**
+ * Contains references to all currently open windows.
+ *
+ * @type {Array}
+ */
+let windows: Electron.BrowserWindow[] = [];
+
+/**
+ * The index of the currently focussed window in `windows`.
+ *
+ * @type {number}
+ */
+let focussedWindowIndex: number;
 
 if (app.dock) {
     app.dock.setIcon(nativeImage.createFromPath("build/icon.png"));
 }
 
 app.on("ready", () => {
+    createWindow();
+    registerApplicationMenu();
+
+    globalShortcut.register('CommandOrControl+N', () => {
+        createWindow();
+    })
+});
+
+app.on("window-all-closed", () => {
+    // On OSX the app usually keeps running until Cmd+Q'd.
+    if (process.platform !== 'darwin') {
+        app.quit();
+    }
+});
+
+app.on("open-file", (_event, file) => {
+    if (focussedWindowIndex > -1) {
+        windows[focussedWindowIndex].webContents.send("change-working-directory", file);
+    }
+});
+
+/**
+ * Create a new terminal window and push it to `windows`.
+ */
+export function createWindow() {
     const bounds = windowBounds();
 
     let options: Electron.BrowserWindowOptions = {
@@ -32,20 +72,65 @@ app.on("ready", () => {
 
     browserWindow.loadURL("file://" + __dirname + "/../views/index.html");
 
-    browserWindow.on("focus", () => app.dock && app.dock.setBadge(""));
+    browserWindow.on("focus", () => {
+        app.dock;
+        app.dock.setBadge("");
+        setFocssedWindow(browserWindow);
+    });
 
     browserWindow.webContents.on("did-finish-load", () => {
         browserWindow.show();
         browserWindow.focus();
     });
 
-    app.on("open-file", (_event, file) => browserWindow.webContents.send("change-working-directory", file));
-});
+    browserWindow.on('closed', () => {
+        closeWindow(browserWindow);
+    })
 
-app.on("window-all-closed", () => app.quit());
+    windows.push(browserWindow);
+}
 
-ipcMain.on("quit", app.quit);
+/**
+ * Remove a window from the `windows` list.
+ *
+ * @param {BrowserWindow} browserWindow the window to remove
+ */
+function closeWindow(browserWindow: Electron.BrowserWindow) {
+    let index = windows.indexOf(browserWindow);
+    windows.splice(index, 1);
 
+    // Focus next window
+    let next_window = Math.min(index, windows.length);
+    if (next_window > -1) {
+        windows[next_window].focus();
+    }
+    else {
+        focussedWindowIndex = -1;
+    }
+}
+
+/**
+ * Set `focussedWindowIndex` to the index of the passed-in BrowserWindow.
+ *
+ * @param {Electron.BrowserWindow} browserWindow
+ */
+function setFocssedWindow(browserWindow: Electron.BrowserWindow) {
+    focussedWindowIndex = windows.indexOf(browserWindow);
+}
+
+/**
+ * Register the global application menu.
+ */
+function registerApplicationMenu() {
+    const template = buildMenuTemplate(app);
+    Menu.setApplicationMenu(Menu.buildFromTemplate(template));
+}
+
+/**
+ * Get the maximum possible window dimensions on the display.
+ *
+ * @return {Electron.Rectangle}
+ */
 function windowBounds(): Electron.Rectangle {
     try {
         return JSON.parse(readFileSync(windowBoundsFilePath).toString());
--- a/src/main/Main.ts
+++ b/src/main/Main.ts
@@ -25,9 +25,7 @@ app.on("ready", () => {
     createWindow();
     registerApplicationMenu();
 
-    globalShortcut.register("CommandOrControl+N", () => {
-        createWindow();
-    });
+    globalShortcut.register("CommandOrControl+N", createWindow);
 });
 
 app.on("window-all-closed", () => {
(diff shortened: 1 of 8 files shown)


jsphpl:
Sorry guys, this has gotten larger than expected.

Had to refactor the menu item handlers to emit events on the passed-in `BrowserWindow` so the commands are directed to the currently focussed window. Hope this won't break with any other changes currently in the queue…

What do you think, @pmkary, @drew-gross & @vshatskyi ?

src/main/Main.ts
```diff
@@ -1,12 +1,52 @@
...
 import {readFileSync} from "fs";
 import {windowBoundsFilePath} from "../utils/Common";
+import {buildMenuTemplate} from "./Menu";
+
+/**
+ * Contains references to all currently open windows.
+ *
+ * @type {Array}
+ */
+let windows: Electron.BrowserWindow[] = [];
```
drew-gross: Do you think it might be possible to have a `Set` of `unfocussedWindows` and a single, optional, `focussedWindow`, with windows moving into and out of the `focussedWindow`? That feels like it might be a little cleaner than an array with an index. I think it would probably also be fine as is if you don't have the time to change it.

jsphpl: I saw the tabs are stored the same way, just in an array with an extra variable containing the index of the currently focussed one. Would be probably clean to keep consistent with the rest of the project, what do you think?

pmkary: @jsphpl I agree with @drew-gross so much. Having `sets` for the job makes it hell lot easier to manage the communications in windows.

src/main/Main.ts
```diff
@@ -32,20 +72,65 @@ app.on("ready", () => {
...
+        focussedWindowIndex = -1;
+    }
+}
+
+/**
+ * Set `focussedWindowIndex` to the index of the passed-in BrowserWindow.
+ *
+ * @param {Electron.BrowserWindow} browserWindow
+ */
+function setFocssedWindow(browserWindow: Electron.BrowserWindow) {
```
jsphpl: Thanks, fixed.

src/main/Main.ts
```diff
@@ -1,12 +1,52 @@
...
 
 if (app.dock) {
     app.dock.setIcon(nativeImage.createFromPath("build/icon.png"));
 }
 
 app.on("ready", () => {
+    createWindow();
+    registerApplicationMenu();
+
+    globalShortcut.register("CommandOrControl+N", () => {
```
jsphpl: Done.

src/main/Main.ts
drew-gross: Maybe we could update tabs to use the `Set`strategy too :p up to you. I don't mind too much either way.

vshatskyi: This data structure is called a zipper: a "traversable" with focus. We can create a polymorphic one and use it everywhere. And array plus an index is a legit array zipper, but we could make a nice interface for it.

src/views/1_ApplicationComponent.tsx
```diff
@@ -9,33 +9,77 @@ import {saveWindowBounds} from "./ViewUtils";
...
+        this.electronWindow.webContents
             .on("devtools-opened", () => this.recalculateDimensions())
             .on("devtools-closed", () => this.recalculateDimensions());
 
-        ipcRenderer.on("change-working-directory", (_event: Electron.IpcRendererEvent, directory: string) =>
-            this.focusedTab.focusedPane.session.directory = directory,
-        );
+        ipcRenderer
+            // Tabs actions
+            .on("change-working-directory", (_event: Electron.IpcRendererEvent, directory: string) =>
```
jsphpl: Good idea! Shall i introduce constants for each in `Enums`?

src/main/Main.ts
jsphpl: The zipper sounds interesting, any further hints or code examples?

jsphpl: Maybe something like this?
```TypeScript
interface Zipper {
	add(item: any);
	delete(item: any);
	focus(item: any);
}
```
This acually is the first time for me to write typescript, so i would probably need some help here. If we work on the interface together, i can write the code.

Didn't quite get the `<T>` concept yet, but it might help us to get rid of the `any`'s, right?

src/main/Main.ts
```diff
@@ -1,12 +1,50 @@
...
+ * Contains references to all currently open windows.
+ *
+ * @type {Array}
+ */
+let windows: Electron.BrowserWindow[] = [];
+
+/**
+ * The index of the currently focussed window in `windows`.
+ *
+ * @type {number}
```
vshatskyi: I don't mind, it's just this information already exists as types.

src/views/1_ApplicationComponent.tsx
vshatskyi: Ideally I'd like to have direct function calls instead of events, but I'm not sure it's possible. An enum would be nice too. Alternatively you can extend the interface of the `on` function to accept particular strings. This way you can verify the second argument based on the first one.

src/main/Main.ts
pmkary: @jsphpl I have a suggestion. Wait I'll be adding it right now..

pmkary: Okay I believe this will do it:

```TypeScript
//
// ─── ZIPPER CLASS ───────────────────────────────────────────────────────────────
//

    /** A _Set_ that has one __active__ element */
    class Zipper extends Set {

        //
        // ─── EXTRA STORAGE ───────────────────────────────────────────────
        //

            private activeElement: any;

        //
        // ─── CONSTRUCTOR ─────────────────────────────────────────────────
        //

            constructor ( ) {
                super( );
                this.activeElement = null;
            }

        //
        // ─── REMOVE ELEMENT ──────────────────────────────────────────────
        //

            public delete ( element ) {
                // if the active index is the index we want
                // to delete we have to change it.
                if ( this.activeElement === element ) {
                    // finding the active index
                    let activeIndex = 0;
                    for ( let index in this.values ) {
                        if ( this.values[ index ] === element ) {
                            this.activeElement = index;
                            break;
                        }
                    }

                    // checking what to do:
                    if ( activeIndex > 0 ) {
                        this.activeElement = this.values[ activeIndex - 1 ];
                    }

                    else if ( activeIndex < this.values.length ) {
                        this.activeElement = this.values[ activeIndex + 1 ];
                    }

                    else {
                        this.activeElement = null;
                    }
                }

                // super
                this.delete( element );
            }

        //
        // ─── SET ACTIVE INDEX ────────────────────────────────────────────
        //

            public activate ( element ) {
                if ( this.has( element ) ) {
                    this.activeElement = element;
                    return true;
                } else {
                    return false
                }
            }

        //
        // ─── GET FOCUSED ELEMENT ─────────────────────────────────────────
        //

            public getActiveElement ( ) {
                return this.activeElement;
            }

        // ─────────────────────────────────────────────────────────────────

    }

// ────────────────────────────────────────────────────────────────────────────────

```

src/main/Main.ts
pmkary: @jsphpl typescript is aware of the params so jsdoc is no longer needed... You can just explain what the parameter does. But you don't need to specify any types on the documentations...

src/main/Main.ts
vshatskyi: @pmkary I'd disallow creating an empty set. This'd eliminate the need to null-check `getActiveElement`. Anyway, it's getting too complex; I'm fine if we implement it separately from multiple windows.

pmkary: @vshatskyi I also need zipper for my apps. So I'm on it. Let me make a full bullet-proof object and then I'll make a PR for that.

vshatskyi: @pmkary, then maybe you'd be interested in a way to restrict the domain of the `activate` function to only accept existing elements. https://github.com/vshatskyi/black-screen/blob/3e24644dd8ec5a16f240e05291437000497588a7/src/shell/Aliases.ts#L44-L50

pmkary: @vshatskyi that's right, I'll be adding that too...

src/views/1_ApplicationComponent.tsx
jsphpl: What about a class with two abstract methods `send()` and `receive()`. Each "message channel"  (e.g. change-working-directory) would be a subclass, where the programmer's task would be to make sure both methods have the same signature.

In the "emitting" process, one would call `send()`, which has the sole purpose of serializing its arguments using private methods and emit the ipc message. The "listening" process would subscribe to the channel by calling a method for that purpose on our class, causing our `receive()` method to be executed on the receipt of the message. The `receive()` call would be made internally by our class, which would take care of subscribing to the channel and deserializing the arguments.

Does that make any sense and would that even be helpful?

One problem i see here is that we'd have to bring other variables into the scope of the `receive()` method…

Here how the implementation for our "change-working-directory" example could look like:
```TypeScript
class ChangeWorkingDirectory extends RemoteCall {
	/**
	 * Only there to call `serializeAndEmit` with all its arguments.
	 */
	send(directory: string) {  // same signature
		this.serializeAndEmit({directory: directory});
	}
	/**
	 * Implement the functionality here.
	 */
	receive(directory: string) {  // same signature
		this.context.focusedTab.focusedPane.session.directory = directory;
		// context could be passed in on event subscription
	}
}
```

src/main/Main.ts
jsphpl: @pmkary is there a typescript pendant of jsdoc that makes use of that fact?

src/main/Main.ts
jsphpl: So what should i do? I'd like to avoid the initial suggestion from @drew-gross (sorry) because it's an additional shifting around of data, especially given that it's gonna be replaced by a zipper. In order to achieve a similar outcome for the meantime, we could just introduce a `getFocussedWindow()` function for ease of access, what do you think?

src/main/Main.ts
pmkary: @jsphpl Most of the codes [out there](https://github.com/Microsoft/vscode/blob/master/src/vs/editor/contrib/folding/common/indentFoldStrategy.ts#L16-L18) simply use plain `/** ... */` comments. But you can have `@param` and stuff too so like:

screen shot 2016-12-11 at 4 37 18 pm

But it's for just when you really need to specify the each param so you can have hover options. Most of the times you just go with a simple description of the function and because TypeScript is aware of the types it generates you a perfectly well explained scheme of the function:

screen shot 2016-12-11 at 4 43 07 pm

Also you can have markdown: 

screen shot 2016-12-11 at 4 50 07 pm

So you can just bold or italic the params and type faster and more clear

vshatskyi:
@jsphpl I plan to close old PRs to keeps things clean. Is it OK with you if I close this one?

jsphpl:
@vshatskyi yes, please close it. Looks like this has fallen asleep and needs a new attempt off the new codebase. Has there been any progress on multi window support in the mean time?

PR #924, opened 2016-12-07
Fix conflicting keybindings for "fullscreen" and "find" on macOS (CMD+F)

jsphpl:
On macOS, `Edit>Find` has the same keybinding as `View>Toogle Full Screen`, namely `CMD+F`.

This pull request reassigns full screen to `CMD+Shift+F` on mac. As a side-effect, the accelerator can be unified to `CmdOrCtrl+Shift+F` which is equal to the previous mapping on all other operating systems.


drew-gross:
Awesome, thanks!

PR #954, opened 2017-01-22
fix autocomplete on linux

sylvaindumont:
On linux autocompletion doesn't work and `npm test` fail on autocompletion because of
`/bin/sh: 1: printf: %q: invalid directive`

Setting child_process default shell to /bin/bash fix autocompletion and tests.
I don't know if it causes any issues on osx, but on linux it works like a charm.

### DIFF
--- a/src/PTY.ts
+++ b/src/PTY.ts
@@ -65,6 +65,7 @@ export function executeCommand(
             ...execOptions,
             env: _.extend({PWD: directory}, process.env),
             cwd: directory,
+            shell: "/bin/bash",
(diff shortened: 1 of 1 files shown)


drew-gross:
I think this is probably good, but I'll let @vshatskyi weigh in if he wants before merging. We try to use the user's default shell currently, but only really have support for `bash` and `zsh`. Maybe the right thing to do here is use whatever the user's default is, but supporting multiple shells in our autocomplete stuff sounds like a pain.

PR #964, opened 2017-01-31
Allow word wrapping in the prompt.

### LINKED ISSUE #963: Long commands cause weird display problems
Reproduce by typing in the command:

> testtestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestesttesttestest

(My real command that triggered this was `make && run_one_test && run_another_test && etc.`)

There are several problems:

1. The cursor goes off the end of the window so you can't see what you're typing
2. You can sort of scroll right by it just makes the window flash weirdly
3. Moving the cursor when it's off the side of the window also makes the window flash weirdly
4. After running the command, you can now scroll the whole window to the right, which moves the current command but ironically doesn't actually move the old command so you can see it.

I think you just need to let commands overflow to the next line when they get too long?

brendanlong:
This requires several changes:

 - Set several containers max-width to 100%
 - Don't force a height on the prompt
 - Allow wrapping when necessary in the prompt (pre-wrap)
 - Force word breaks when necessary in the prompt

Fixes #963

### DIFF
--- a/src/views/4_PromptComponent.tsx
+++ b/src/views/4_PromptComponent.tsx
@@ -150,7 +150,7 @@ export class PromptComponent extends React.Component<Props, State> {
                 </div>
                 <div
                     className="prompt" // Used by tests
-                    style={css.prompt(this.state.isSticky)}
+                    style={css.prompt}
                     onInput={this.handleInput.bind(this)}
                     onDrop={this.handleDrop.bind(this)}
                     onBlur={() => this.setState({...this.state, caretPositionFromPreviousFocus: getCaretPosition(this.commandNode)})}
--- a/src/views/css/main.ts
+++ b/src/views/css/main.ts
@@ -74,6 +74,7 @@ const applicationGrid = {
     },
     sessions: {
         height: "100%",
+        maxWidth: "100%",
     },
 };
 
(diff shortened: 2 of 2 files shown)


brendanlong:
An alternate solution would be allowing the prompt to scroll horizontally, but that seems less nice?

drew-gross:
I like this better than allowing scroll. Will merge once CI passes.

drew-gross:
Slightly concerned about the removal of `isSticky` though. Does this still look OK for prompts that have been scrolled to the top? Could you include a screenshot?

brendanlong:
Ah so it looks good with one thing stuck to the top but starts to look bad if more stuff scrolls on top of it:

screen shot 2017-02-01 at 9 18 09 am

screen shot 2017-02-01 at 9 18 23 am

I guess we probably need to force sticky prompts to size down once something else scrolls on top of them..

drew-gross:
I think it would also be reasonable to just make it always size down when it's that the top if that's easier to implement.

r24y:
Would it be possible to use `text-overflow: ellipsis` for sticky prompts?

vshatskyi:
@r24y I plan to close old PRs to keeps things clean. Is it OK with you if I close this one?

brendanlong:
I'm fine with closing this; I doubt I'll have time to make more changes to this.

Issue #957, opened 2017-01-27
Aliases with quotes, single or double, aren't working


ainzzorl:
These commands fail:
alias l="ls -l"
alias l='ls -l'
alias l="ls"
alias l='ls'

This succeeds:
alias l=ls

I believe it treats l="ls -l" as 2 arguments: l= and "ls -l", but they should be one.

drew-gross:
Are you talking about in your `.bashrc`?

ainzzorl:
No: see the screenshot.

![screen shot 2017-01-27 at 9 36 03 am](https://cloud.githubusercontent.com/assets/25125937/22380926/40824b06-e474-11e6-9d9d-550a7a26e541.png)

drew-gross:
Cool, thanks for the update.

ainzzorl:
I'd like to try to fix this.

I think the root cause is a bug in the Scanner: every quoted string literal is a separate token, but it shouldn't be. Example: echo a"b"'c'. Expected output: abc, actual: a b c. The simplest way to fix it is to do this: if there are 2 string literals next to each other with no space characters between them, then combine them into one. Does this make sense?

drew-gross:
Yeah, that does. The scanner code always did look a little suspicious to me. We would definitely appreciate the help! (I would also appreciate a test case in `scanner_spec.ts` if you do fix this)

PR #966, opened 2017-02-08
Fix scanner splitting literals with quotes (#957)

ainzzorl:
Fixes https://github.com/vshatskyi/black-screen/issues/957

The problem is that Scanner splits tokens made up of several string literals, like key="value". The solution is to squash literals into composite literals if there are no spaces between them.

I moved StringLiteral class because it needs to be above Word class, because Word now extends StringLiteral.

### DIFF
--- a/src/shell/Scanner.ts
+++ b/src/shell/Scanner.ts
@@ -31,7 +31,13 @@ export class Empty extends Token {
     }
 }
 
-export class Word extends Token {
+export abstract class StringLiteral extends Token {
+    get value() {
+        return this.raw.trim().slice(1, -1);
+    }
+}
+
+export class Word extends StringLiteral {
     get value() {
         return this.raw.trim().replace(/\\\s/g, " ");
     }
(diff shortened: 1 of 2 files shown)


drew-gross:
Looks like this works, thanks for adding tests. Is there no way to do this in the `patterns` section? Like maybe allowing quotes in a `Word` then making sure `*QuotedStringLiteral` requires whitespace at the start (or start of string anchor). If we could do that, it would probably be a lot simpler.

ainzzorl:
I can try writing it this way. But I doubt it'll turn out better because:
1. If it allows quotes in a Word, then it should allow spaces too, but only inside quotes. The Word regex is rather complex already, and extending it to support spaces inside quotes will make it really hard to understand.
2. We'll need to override Word.value and Word.escapedValue to not split text inside quotes. It shouldn't be much simpler than what's in this PR.
3. What to do if there's an odd number of quotes in the string? However, neither the code in this PR nor what's in master handles it correctly, so it should be considered a different issue.

drew-gross:
Alright, I believe you. Merging.

PR #969, opened 2017-02-19
Use "git symbolic-ref" to determine active Git branch

mkenne11:
Modified the "branches" function to use the "git symbolic-ref --short -q HEAD" command to determine the active Git branch.

### DIFF
--- a/src/utils/Git.ts
+++ b/src/utils/Git.ts
@@ -2,6 +2,7 @@ import {linedOutputOf} from "../PTY";
 import * as Path from "path";
 import * as fs from "fs";
 import * as _ from "lodash";
+import {executeCommand} from "../PTY";
 
 export class Branch {
     constructor(private refName: string, private _isCurrent: boolean) {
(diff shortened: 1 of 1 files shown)


drew-gross:
Awesome, thanks for the help!

mkenne11:
No worries at all happy to help!

I'm not super on top of asynchronous calls in ts/js. See my question inline about the two async calls which "lines.map" is dependent on.

src/utils/Git.ts
```diff
@@ -121,12 +122,20 @@ export async function branches({
...
+        ['"symbolic-ref"', '"--short"', '"-q"', '"HEAD"'],
+        directory)
+        .then(branchName => branchName.trim());
     let lines = await linedOutputOf(
         "git",
         ["for-each-ref", tags ? "refs/tags" : "", "refs/heads", remotes ? "refs/remotes" : "", "--format='%(HEAD)%(refname:short)'"],
         directory,
     );
-    return lines.map(line => new Branch(line.slice(1), line[0] === "*"));
+    return lines.map(line => {
```
mkenne11: Can I be sure that both the "currentBranch" and "lines" asynchronous calls always finish before "lines.map" code runs?

drew-gross: Yep. The `await` guarantees that.

mkenne11: Thanks for the response.

PR #991, opened 2017-03-06
Fixes #978

### LINKED ISSUE #978: "Toggle Full Screen" and "Find" keybindings are identical
The keybindings/shortcuts for both "Toggle Full Screen" and "Find" are identically <kbd>⌘F</kbd>. When I try to "Find" something, it only toggles full screen mode.

In either case, the "Toggle Full Screen" command should be <kbd>⌃⌘F</kbd> as it is in all native macOS apps.

williamtdr:
Changes the keybinding for toggling full screen to match MacOS conventions, as found in other apps.


drew-gross:
Thanks for the help! Closing because we got a different PR that does the same thing.

PR #996, opened 2017-03-12
Support combinations of short flags for auto-completion providers

mkenne11:
Added function 'combineShortFlags' to support entry of short flag combinations for auto-complete providers.

### DIFF
--- a/src/plugins/autocompletion_utils/Common.ts
+++ b/src/plugins/autocompletion_utils/Common.ts
@@ -386,3 +386,24 @@ export const commandWithSubcommands = (subCommands: SubcommandConfig[]) => {
         return [];
     };
 };
+
+export const combineShortFlags = (suggestionsProvider: AutocompletionProvider) => {
+    return async (context: AutocompletionContext) => {
+        const token = context.argument.value;
+        const reShortFlag = new RegExp(/^\-[a-zA-Z]$/);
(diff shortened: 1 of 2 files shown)


mkenne11:
This is a possible implementation to address #596.

src/plugins/autocompletion_utils/Common.ts
```diff
@@ -386,3 +386,24 @@ export const commandWithSubcommands = (subCommands: SubcommandConfig[]) => {
...
+    return async (context: AutocompletionContext) => {
+        const token = context.argument.value;
+        const reShortFlag = new RegExp(/^\-[a-zA-Z]$/);
+        const reShortFlags = new RegExp(/^\-[a-zA-Z]+$/);
+        const suggestions = await suggestionsProvider(context);
+        if (reShortFlags.test(token)) {
+            return suggestions
+                    .filter(s =>
+                        reShortFlag.test(s.value) && !token.includes(s.value.slice(1))
+                            && s.withSpace)
```
mkenne11: I am assuming the (suggestion) `withSpace` attribute is used for short flag options with values, so these should not be included when suggesting combined short flag auto-complete options.

drew-gross: That is correct

drew-gross:
This looks awesome! Thanks for the help!

Issue #977, opened 2017-03-02
Rotating active pane should wrap around


akivajgordon:
## Context
Suppose I have, say, 3 panes open, and I am advancing to the next pane (like with <kbd>cmd-J</kbd>). For this context, let's say, for example, I'm currently on the last pane.

## Current behavior
Nothing happens when I press <kbd>cmd-J</kbd>.

## Desired behavior
When I press <kbd>cmd-J</kbd>, I should now be on the 1st pane. And similarly, the reverse should be true if try to go to the previous pane when currently on the 1st pane – I should end up on the last one. Basically, it should wrap around.

This would fall in line with how tabs in Black Screen seem to be implemented.

mkenne11:
@drew-gross can this be closed now?

drew-gross:
Yep. Thanks!

PR #1001, opened 2017-03-24
Allow wrap around when advancing next & previous pane

mkenne11:
Update implements wrap around when advancing next & previous panes.

Code to address issue #977.

### DIFF
--- a/src/utils/PaneTree.ts
+++ b/src/utils/PaneTree.ts
@@ -78,6 +78,8 @@ export abstract class PaneList {
     }
 
     previous(pane: Pane): Pane {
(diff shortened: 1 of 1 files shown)


mkenne11:
I added jsdoc comments for 'next' and 'previous' methods. I noticed the rest of the project doesn't have these - should I leave them in?

drew-gross:
Thank you! We don't use jsdoc right now, but there is no harm in leaving them there. I think once we have a public API for plugins, jsdoc might come in handy.

mkenne11:
No worries.

PR #1002, opened 2017-03-26
Improve performance of PaneList next & previous methods

mkenne11:
Simplified code and improved performance of PaneList next & previous methods.


drew-gross:
👍

PR #1025, opened 2017-05-15
Updated vscode debugging instructions

mkenne11:
Made some updates to the vscode debugging instructions to:
- Use the node "inspector" debug protocol.
- Removed erroneous instructions for debugging compiled javascript file.
- Changed vscode tested to 1.12.1.


mkenne11:
Using node "inspector" debug protocol allows debugging across asynchronous calls. Also the previous protocol "legacy" setting will not be supported for node v8.0 and later.

Issue #1021, opened 2017-05-10
'git branch' list items have directory names cut off when using "/"


pacificpelican:
Sort of an edge case maybe, but at my work we have adopted the (anti-)pattern of using directory names in Git branches, so for example if I am working on bugfixes I might call my branch <code>Dan/bugfixesMay2017</code> or whatever.  Unfortunately when I type <code>git branch</code> in Black Screen it only shows the last part of the name [<code>bugfixesMay2017</code>]

mkenne11:
I was able to replicate this bug. A branch I created was named ```feature/test1```. See screenshot below after running  ```git branches``` in bash terminal showing this branch.
![image](https://cloud.githubusercontent.com/assets/12181805/26449265/9f4b6dfc-4194-11e7-83d5-76187a45fafb.png)

When running the ```git branch``` command in black-screen it shows the ```feature/test1``` branch name truncated to ```test1```. See following screenshot.
![image](https://cloud.githubusercontent.com/assets/12181805/26449366/1f83c7f8-4195-11e7-8a2e-8f081b0e0b6c.png)

mkenne11:
This bug is caused by the ```Branch``` class constructor truncating the branch-name using "/". The "/" character is a valid character in branch names - so it's not always the case that a "/" separates a ref-remotes, ref-tags or ref-heads name from a branch name.

mkenne11:
See suggested fix in PR #1041. The change generates the full branch names including ref-heads, ref-remotes and ref-tags prefixes; and using a regex filter removes these prefixes.

See black-screen screenshot with branch suggestions showing ```feature/test1``` correctly.
![image](https://cloud.githubusercontent.com/assets/12181805/26450112/203814de-4199-11e7-831e-9d83242f10f9.png)

PR #1041, opened 2017-05-25
Fix issue #1021. Split branch name correctly from ref-heads/remotes/tags

mkenne11:
Fixed issue #1021. 'Branch' class constructor incorrectly split branch-name from ref-heads, ref-remotes and ref-tags names using the "/" character.

The "/" character is valid in branch names.

### DIFF
--- a/src/utils/Git.ts
+++ b/src/utils/Git.ts
@@ -1,7 +1,6 @@
 import {linedOutputOf} from "../PTY";
 import * as Path from "path";
 import * as fs from "fs";
-import * as _ from "lodash";
 import {executeCommand} from "../PTY";
 
 export class Branch {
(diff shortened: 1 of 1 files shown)


mkenne11:
Failed to build - not sure why.

src/utils/Git.ts
```diff
@@ -127,16 +126,22 @@ export async function branches({
         ['"symbolic-ref"', '"--short"', '"-q"', '"HEAD"'],
         directory)
         .then(branchName => branchName.trim());
+    const branchFilter = [
+        "s/^refs\\/heads\\///g",
+        remotes ? "s/^refs\\/remotes\\/[^\\?%\\*\\:\\|\"\\<\\>\\\\]*\\///g" : "",
+        tags ? "s/^refs\\/tags\\///g" : ""].join(";");
```
vshatskyi: This seems really complicated. Can we do two calls, one for heads and tags with `--format='%(HEAD)%(refname:strip=2)'`, and one for remotes with `strip=3`, as it's done in bash completion? https://github.com/git/git/blob/master/contrib/completion/git-completion.bash

PR #1042, opened 2017-05-25
Fix issue #1021. Split branch name correctly from ref-heads/remotes/tags

mkenne11:
Fixed issue #1021. 'Branch' class constructor incorrectly split branch-name from ref-heads, ref-remotes and ref-tags names using the "/" character.

The "/" character is valid in branch names.


mkenne11:
CI build appears to pass unit tests, linter and compile. However it failed on application launch.
image

vshatskyi:
@mkenne11 could you please add a test for that? I'm experiencing the same CI errors; will try to fix.

src/utils/Git.ts
```diff
@@ -127,16 +126,22 @@ export async function branches({
         ['"symbolic-ref"', '"--short"', '"-q"', '"HEAD"'],
         directory)
         .then(branchName => branchName.trim());
+    const branchFilter = [
+        "s/^refs\\/heads\\///g",
+        remotes ? "s/^refs\\/remotes\\/[^\\?%\\*\\:\\|\"\\<\\>\\\\]*\\///g" : "",
+        tags ? "s/^refs\\/tags\\///g" : ""].join(";");
```
vshatskyi: This seems really complicated. Can we do two calls, one for heads and tags with `--format='%(HEAD)%(refname:strip=2)'`, and one for remotes with `strip=3`, as it's done in bash completion? https://github.com/git/git/blob/master/contrib/completion/git-completion.bash

vshatskyi:
@mkenne11 for your information, tests are now passing in master.

src/utils/Git.ts
mkenne11: Thanks good tip! I've made some code updates but since the repo rename to "upterm" and some other commits branch auto-complete suggestions for "git checkout" don't seem to show for me now. I'll have another look in the next few days.

mkenne11:
Thanks @vshatskyi I made code updates as suggested and the PR now builds without errors. On my local machine branch suggestions aren't showing for "git checkout" since I rebased recent changes made. I'll have another look over the next few days.

mkenne11:
You asked if I could write some unit tests. I plan too, however this may take some time. I am new to mocha and chai, and also need to setup travis ci on my local machine. 
Did you want to wait until I write unit tests before accepting the PR?

vshatskyi:
@mkenne11 I'll merge it as is, thanks for fixing it!

mkenne11:
No worries - thanks!

Issue #800, opened 2016-10-05
Windows


di48l069:
Hi,

Please bring support for Windows including a downloadable installer

drew-gross:
Happy to accept help :) I don't think any of the current core contributors use Windows so don't expect this to be high priority for them.

gyurisc:
I would like to help bringing black-screen to windows, but I am not sure where to start.

minecrawler:
@gyurisc

start by downloading the source, installing all dependencies and running it (it's just Electron, so no need for any fancy compilation for now). Just getting it started without errors is your first hurdle ;)

Once it runs, decide on a basic feature set you think has to work no matter what. Then see if those features work. If they don't, debug the application and fix them in a way which does not prevent them from working on other platforms.

As soon as all core features work, you should make a PR with your fixes, package Electron and the application into a binary bundle (ZIP file should be enough for now, I would leave the decision which installer to use etc. for the core team). Then you can send the ZIP archive to the core team and have them include it with the release you based it on (just add it as a binary file in the RELEASE section)

You might want to ask them to add a direct link to the file to the README, so people will find it more easily.

From there on, you will have to make sure all subsequent releases work on Windows. Also try to add more working features.

jsphpl:
Duplicate of #302

LogoiLab:
I find it annoying that "just because our contributors don't use it we shouldn't support it" is your response. You should aim for cross platform from the start. Also, look at the percentage of the world population that actually owns a Mac. It's JavaScript guys c'mon, it's been cross platform since around '98 has it not?

weaversam8:
Indeed, it is annoying, but part of maintaining open source software is the fact that people are volunteering their own time and resources to make it. Those things are limited, and while I personally love Windows and despise Macs, if the contributors have everything they need then it may not be high on their priorities. That being said, this is open source after all, and you're welcome to try and make the port yourself. @minecrawler provided some great ideas on how to start.

drew-gross:
Well put @weaversam8. I do think we should support Windows, but I have typically 2-3 hours a week to work on black-screen, and I'm going to put those hours towards things that are important to me, first and foremost.

Also, I think it should already pretty much work on Windows, maybe with some small updates to the build system, but I don't have a Windows machine, and it's not easy to test Windows in CI. If someone who does have Windows and knows a little about Electron tried to add support I doubt it would be too hard.

weaversam8:
[...]

PR #1061, opened 2017-05-28
Change README.md explaining unofficial Windows Support

weaversam8:
I'm not sure if it's a good time to say this, but seeing as there is now experimental support (see #800) we should potentially update the README.md explaining that it's theoretically possible.


vshatskyi:
Thanks, @weaversam8!

Issue #1066, opened 2017-05-30
Git status bar only works in repository root


sombrelain:
The git repository information (branch, changes, etc.) shown in the status bar are only visible when you are in the root directory of the repository. Not sure if this is intended, but I would expect them to also show up in subdirectories of a repository.

One possible fix would be to adjust the `isGitRepository` in the git util to also recursively check parent directories for existing `.git` folders (up to root). I made the change locally and it seems to work fine, unfortunately can't test it on windows right now. 

If no one objects, I'd just make a pull request out of it?

vlad-shatskyi:
Yeah, you can go ahead and open a PR.

vlad-shatskyi:
Fixed.

PR #1078, opened 2017-06-06
Make git statusbar also work in repository sub-directories

sombrelain:
The git util will now recursively go up through directories when trying to find a .git subfolder. It will stop when it either finds one, or reaches the filesystem root. I don't have a windows box available to test, according to the documentation `path.root` should work though.

I'm somewhat new to TS, if there's any codewise feedback, I'm happy to receive.

This fixes issue #1066


vlad-shatskyi:
@sombrelain please use https://stackoverflow.com/a/16925062/1149074

sombrelain:
I've been thinking about that already, but was not exactly sure how to handle the type guard. The method needs to be asynchronous imho, but that does no longer allow it to work as type guard (or maybe it does, but that's beyond my TS knowledge and I couldn't find anything on it either). I made the changes anyhow, I'm just not sure if the intersection type is still needed at all.

If the solution is not optimal, I'm happy to do some more research and adjust the code.

sombrelain:
Hmm, I think I need to put some more thought in that. I'll close the pull request and start a new one when i have a clean solution, just so this does not get too messy.

PR #1086, opened 2017-06-09
enable ⌘+k to clear terminal

preschian:


MindTooth:
Seems that you added your whole `package.lock` file.  Intended?

vlad-shatskyi:
While I understand that it's the default shortcut in iTerm, there is already one for clearing previous output - `Ctrl+L`.

In the future they'll be configurable, but for now I'll decline this PR. Sorry about that.

PR #1087, opened 2017-06-09
open in browser/finder from statusbar

preschian:
never using typescript before, sorry for my bad code


vlad-shatskyi:
I'm sorry, @preschian, but I can't accept this feature because I don't think opening a GitHub page is what people expect when they click on the branch name. It would be better if you opened an issue first, and only implemented it if we agree it's worth implementing.

I understand it's useful for *you*, and maybe in the future we'll have an API allowing such kinds of extensions.

PR #1143, opened 2017-07-30
Change close button cursor to pointer in TabHeader

sorousht:
Just add one line of css to .close-button in index.html

### DIFF
--- a/src/views/index.html
+++ b/src/views/index.html
@@ -115,6 +115,7 @@
             position: absolute;
             left: var(--margin);
             top: calc(var(--margin) / 2);
+            cursor: pointer;
         }
 
         .tab-header:hover .close-button {


vlad-shatskyi:
Hey, @sorousht. I'm reluctant to accept this pull request because

* I think a cross pictogram already has strong affordance https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b
* Other applications have the default cursor for this action (e.g. Google Chrome).

But I'd like to hear your reasoning for this change.

sorousht:
Hey @vlad-shatskyi. Thanks for your response. I read the article and found that the button shouldn’t have a hand cursor. Then, I suggest using `default` cursor instead of `text` cursor (which is the default for a span element).

vlad-shatskyi:
Hmm, does it look like a vertical bar for you, @sorousht? It looks like an arrow on macOS and Ubuntu.

sorousht:
@vlad-shatskyi Yes, but I see this on Ubuntu!

vlad-shatskyi:
Oh, you're right, I was hovering a wrong element. Alright, I'll merge it as is. Thank you again!

PR #1194, opened 2017-09-03
Changes to ps autocomplete for monaco

mkenne11:
Changes made to the ps autocomplete provider to work with monaco based Suggestion items.

### DIFF
--- a/src/plugins/autocompletion_providers/Ps.ts
+++ b/src/plugins/autocompletion_providers/Ps.ts
@@ -1,34 +1,31 @@
 import {PluginManager} from "../../PluginManager";
-import {
-    shortFlag, mapSuggestions, provide, Suggestion,
-}
-    from "../autocompletion_utils/Common";
 import {combine} from "../autocompletion_utils/Combine";
+import {provide, Suggestion} from "../autocompletion_utils/Common";
 import {AutocompletionContext, AutocompletionProvider} from "../../Interfaces";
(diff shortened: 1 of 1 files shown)


mkenne11:
To show how ps auto-complete now works with this PR I have attached a animated screenshot capture. The capture shows a sample command with multiple option values selected.

vlad-shatskyi:
This is awesome, @mkenne11, thanks!

mkenne11:
No worries thanks @vlad-shatskyi

vlad-shatskyi:
@mkenne11 it fails on my computer (macOS 10.12) with
```
ps: egid: keyword not found
ps: egroup: keyword not found
```

Although, works with these changes:

image

Should I apply them?

mkenne11:
Sorry @vlad-shatskyi I just saw your msg. Yes please apply. Tbh I'm not familiar with the difference between  mac os (and other BSD-based ditributions) vs Linux.
Looking at the ```ps``` man pages for Linux and BSD, BSD ```ps -eo comm``` appears to be equiv to ```ps -eo cmd``` in Linux.

mkenne11:
It could be awkward though - having to check the type if Linux or BSD is running each time before you call this method. Is the OS type easy to check?

vlad-shatskyi:
@mkenne11, it doesn't require IO to check platform. Although, there is more to fix than I originally thought, so I'll leave it as is for now.

PR #1200, opened 2017-09-10
Vscode debug doc changes for config files

mkenne11:
Changes made to vscode debug instructions including:
- Updated "tasks.json" to version 2.0.0 format.
- Compile task references the default "compile" script from package.json.
- Updated "launch.json" to specify Windows path for "runtimeExecutable" parameters.


mkenne11:
@vlad-shatskyi btw do you think it's worth adding both these files under a ".vscode/" folder in the branch (and remove them from .gitignore)? I can see that not everyone uses vscode so maybe not.

vlad-shatskyi:
@mkenne11 I don't think it'd be useful for anyone else, taking into account that nobody really contributes these days. https://github.com/railsware/upterm/graphs/contributors

But if it's useful for you, go ahead and add it.

vlad-shatskyi:
By the way, I didn't ask your feedback, right? Would you mind answering the following questions?

* Do you use Upterm as your main terminal emulator?
* What feature is the best for you?
* Is there anything that annoys you?
* What features do you miss?
* How did you find Upterm?

mkenne11:
Not a problem @vlad-shatskyi see my answers below:

- Do you use Upterm as your main terminal emulator?
No atm I use Ubuntu's standard terminal application.

- What feature is the best for you?
I like auto-suggestion of commands and options. I'm a part-time linux person so this is very helpful.

- Is there anything that annoys you?
The major annoyance is I doesn't keep the sudo session active past one command entry - re-entering my command each time i sudo isn't ideal.
There are also some minor formatting and performance issues as well.

- What features do you miss?
In terms of a wish list of features I'd like to see:
+ Tabs which can be dragged to change there order.
+ A "history" tab which shows previously used tabs for the session, and could be filtered to see all commands or per tab; and the ability to select commands from the history list to re-run.

- How did you find Upterm?
Through the awesome-electron list (https://github.com/sindresorhus/awesome-electron). I was looking for an electron + react + typescript project to contribute too to learn these skills.

Issue #1203, opened 2017-09-15
Commands not available shown in auto-complete provider


mkenne11:
I am running upterm on Ubuntu and a few command auto-complete providers are registered for commands that don't exist on my machine. These commands include - ```brew```, ```rails``` & ```vagrant```.

This results in these commands being suggested. However the linter does pickup the command doesn't exist before I try running it.

PR #1204, opened 2017-09-15
Check command exists before registering auto-complete provider

mkenne11:
Check that a command exists before adding it the command auto-complete provider list.

Possible fix for issue #1203.

### DIFF
--- a/src/PluginManager.ts
+++ b/src/PluginManager.ts
@@ -1,6 +1,6 @@
 import {Prettyfier, EnvironmentObserverPlugin, AutocompletionProvider} from "./Interfaces";
 import * as Path from "path";
-import {io} from "./utils/Common";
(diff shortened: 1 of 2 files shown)


vlad-shatskyi:
I'm afraid this isn't correct implementation. Command existence can change at runtime when you modify $PATH or install new commands. I was thinking of a different solution: lazy load completion providers as a person actually uses some binary name. E.g. `Git.ts` would only be loaded after I type `git`.

mkenne11:
No worries - @vlad-shatskyi I can see your point. I'm not sure how to code that unfortunately.

vlad-shatskyi:
@mkenne11, yeah, it's rather a big lift. I'll do it at some point, but not right now.

PR #1218, opened 2017-09-21
Delegate cmd-f to search component.

### LINKED ISSUE #1180: cmd+f opens find in monaco, instead of switching focus to the search field.

rebornix:
Fix #1180. 

Right now we don't have a way to disable a keybinding and still populate key events in Monaco (a good feature request though I admit). So here I use the same way as Menu to trigger focus on Search Component.


vlad-shatskyi:
I'm afraid it's not how it should work. We just need to unbind cmd+f for monaco.

vlad-shatskyi:
I fixed it in master; closing this PR.

PR #1223, opened 2017-09-21
Fix #1081. Refocus prompt when escape from search component.

### LINKED ISSUE #1081: Prompt isn't focused when the search field loses focus

rebornix:
Search is focused and we need to give the focus back to prompt.


vlad-shatskyi:
Good job, @rebornix, thanks!

vlad-shatskyi:
I just tested it, and prompt does get focus, but typing doesn't work for some reason. Is it the same for you?

PR #1232, opened 2017-09-23
Add autocompletion for commands find, scp, tail

nicolashardy:
Add simple autocompletion for basic commands


vlad-shatskyi:
Thanks, @nicolashardy.

PR #1248, opened 2017-10-02
Symbolic links suggested as directories for cd command

Cygnusfear:
fix for issues https://github.com/railsware/upterm/issues/1014 & https://github.com/railsware/upterm/issues/1225


le0pard:
@Cygnusfear thanks. Looks like fix problem with `ln -s`, but mac alias folders still not visible. Will check, but maybe you will find reason.

le0pard:
Ok, problem found

https://stackoverflow.com/questions/33337057/finder-aliases-not-recognized-as-symlinks-on-a-mac-with-node-js

PR #1278, opened 2017-12-17
Custom inset tittle bar

lSindra:
Hidden the original title-bar and created a new one, with custom buttons.

Not tested in a mac.

### DIFF
--- a/src/views/NavButtonsComponent.tsx
+++ b/src/views/NavButtonsComponent.tsx
@@ -1,0 +1,23 @@
+import * as React from "react";
+
+
+export interface Props {
+    minimize: React.EventHandler<React.MouseEvent<HTMLSpanElement>>;
+    maximize: React.EventHandler<React.MouseEvent<HTMLSpanElement>>;
+    close: React.EventHandler<React.MouseEvent<HTMLSpanElement>>;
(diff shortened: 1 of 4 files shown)


src/views/NavButtonsComponent.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+
+export interface Props {
+    minimize: React.EventHandler<React.MouseEvent<HTMLSpanElement>>;
+    maximize: React.EventHandler<React.MouseEvent<HTMLSpanElement>>;
+    close: React.EventHandler<React.MouseEvent<HTMLSpanElement>>;
+}
+
+export class NavButtonsComponent extends React.Component<Props, {}> {
+    constructor() {
```
le0pard: Just remove this empty `constructor `

src/views/NavButtonsComponent.tsx
```diff
@@ -0,0 +1,24 @@
...
+        super();
+    }
+    render() {
+        // button space not right (big square, small text)
+        return (
+            <div className="navButtons">
```
le0pard: On Mac OS looks like this buttons need to be at left side.

PR #1302, opened 2018-06-02
Fix bug and linter.

ghosind:


src/views/OutputComponent.tsx
```diff
@@ -11,7 +11,7 @@ const CharGroupComponent = ({group}: {group: Char[]}) => {
     return (
         <span
             className="char-group"
-            style={css.charGroup(attributes)}>
+            style={attributes ? css.charGroup(attributes) : null}>
```
le0pard: How about `attributes && css.charGroup(attributes)` or this will not work here?

ghosind: `attributes && css.charGroup(attributes)` is work.

le0pard:
@ghosind thanks

Issue #1240, opened 2017-09-25
Add rake completion


vlad-shatskyi:
Only default tasks for now

PR #1310, opened 2018-07-24
Added basic rake autocompletion

MarmiteCakes:
This commit adds basic completion for rake based on its man page (issue #1240). 
Added an optional parameter to manPageToOptions in utils/ManPages to allow for optional selection of the section to look under for options (rake's are under OPTIONS rather that DESCRIPTION). This could be done manually, but it may be useful for other software, too.


le0pard:
Thanks for PR. You need to fix lint errors (`npm test`)

Issue #1309, opened 2018-07-03
Multi-line pasting is broken


drusepth:
### Steps to reproduce:

1. Paste the following code into the terminal
```
echo "Hello"
echo "World"
```

### Observed behavior:
The lines are concatenated together. What gets run is `echo "Hello" echo "World"`.

### Expected behavior:
The lines are run in sequence, with the first line run first and then the second line run after the first has completed.

glyif:
@drusepth took a look, it looks like both the scanner and the parser are not taking into consideration the new line character.

@le0pard i would say this is a pretty big issue, thoughts?

i can take this bug, will get me more acquainted with upterm's shell. i would think it's using some type of operator-precedence parser for executing commands.

le0pard:
@glyif I am ok, if you want to fix this. Sorry, but don't know, where exactly you need to start.

alex-d-bondarev:
This issue becomes annoying when you copy/paste long commands, like Docker UP with parameters separated by `\` and a new line symbol.
IMHO critical

dwkerwin:
Any update on this?  Just tried upterm for the first time, loved it immediately, then tried to copy/paste a multi-line command and was let down to see it not work.  Pretty much renders it unusable and seems like it would be an easy enough thing to resolve (says the guy complaining but not offering to fix it).

le0pard:
https://github.com/railsware/upterm/issues/1301#issue-327003344

PR #1313, opened 2018-08-29
Fix for Multi-line pasting is broken #1309

ABouzo:
This should fix the issue for multi-line pasting, at least for multiple command. For something like functions would need to be added to the parser.

### DIFF
--- a/src/shell/Parser.ts
+++ b/src/shell/Parser.ts
@@ -21,6 +21,8 @@ export abstract class ASTNode {
     abstract get fullEnd(): number;
 }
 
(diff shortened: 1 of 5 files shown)


le0pard:
Thanks @ABouzo. Does PR ready for review? :)

ABouzo:
Yes, it is. I'm not sure what the process is to mark this as ready for review.

le0pard:
Don't worry. I just was not sure, that came more commits. I will try to review it tomorrow.

ABouzo:
Thank you!

le0pard:
@ABouzo big thanks!

PR #1314, opened 2018-09-14
Tab Headers Legibility Improvements

kevinschaich:
Making the active tab a bit more apparent with borders and background color.

Before:

screen shot 2018-09-14 at 12 31 13 pm

After:

screen shot 2018-09-14 at 12 26 45 pm

Fixes item 3 in https://github.com/railsware/upterm/issues/1192.

### DIFF
--- a/src/views/index.html
+++ b/src/views/index.html
@@ -127,12 +127,12 @@
                 flex-grow: 1;
                 display: flex;
                 text-align: center;
-                border: 1px solid #191C22;
+                border: 1px solid rgba(255, 255, 255, 0.2);
                 font-size: 0.852em;
                 border-top: 0;
-                border-bottom: 0;
                 align-items: center;
                 justify-content: center;
+                border-radius: 0 0 5px 5px;
             }
 
             .tab-header[data-focused=true] {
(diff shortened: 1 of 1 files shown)


src/views/index.html
```diff
@@ -127,12 +127,12 @@
                 flex-grow: 1;
                 display: flex;
                 text-align: center;
-                border: 1px solid #191C22;
+                border: 1px solid rgba(255, 255, 255, 0.2);
```
ABouzo: Why use rgba over hex (#FFFFFF33)? I guess it doesn't matter which is used, but it should be standard across the file for consistency. That's something we can do later on.

kevinschaich: Hex codes don't support transparency. Regardless of if the tab background color changes, this will always make the border "slightly lighter" to stand out.

ABouzo: Huh. I thought you could do `#RRGGBBAA`

kevinschaich: !!! You are right @ABouzo, my mistake. Learn something new every day 😄

kevinschaich:
@ABouzo need a +1 and then I can merge!

le0pard:
@kevinschaich big thanks

PR #1317, opened 2018-10-28
Fix source built in command

eduardomanrique:
I created a fix for source. Maybe it is not the best approach but it is a simple one.


le0pard:
@eduardomanrique thanks for PR. Looks like CI broken.

PR #1322, opened 2019-03-09
Change engines field

zautumnz:
I tested this on 10 and 11. 10 is current LTS, and 6 and 8 aren't getting regular updates, but I left 8 in there because that's what's in the nvmrc.


le0pard:
Big thanks @zacanger

